PR #1, opened 2018-04-01
Flexlayout karma raj

rajgthub:
This is my answer for the homework. Please check and let me know if I need to do anything (@LucyMac ). I have  not considered responsive design at this moment.

### DIFF
--- a/css/_abstract.scss
+++ b/css/_abstract.scss
@@ -8,37 +8,3 @@ $blk-col:#000000;
  justify-content: center;
  align-items: center;
 }
-// MEDIA QUERY MANAGER
-/*
-phone 0 - 600px;
-tab-port < 900px;
-tab-land < 1200px;
-normal < 1800px;
-big-desk > 1800px
-//default 1rem = 16px; 
-inside media query rem and em are same, set to browser defaule
-but research suggests em is better than rem in media query;
-*/
-@mixin respond ($break-point){
-    @if $break-point == phone {
-        @media (max-width: 37.5em) {//600px/16 
-            @content
-        };
-    }
-    @if $break-point == tab-port {
-        @media (max-width: 56.25em) {//900px/16
-            @content
-        };
-    }
-    @if $break-point == tab-land {
-        @media (max-width: 75em) {//1200px/16
-            @content
-        };
-    }
-    @if $break-point == big-desk {
-        @media (min-width: 112.5em) {//1800px/16
-            @content
-        };
-    }
-
-}
--- a/css/_abstract.scss
+++ b/css/_abstract.scss
@@ -1,0 +1,10 @@
+// Colors
+$primary-col:#ff4500;
+$main-left-col:#800080;
+$blk-col:#000000;
+//mixins
+@mixin flx-center{
+ display: flex;
+ justify-content: center;
+ align-items: center;
+}
--- a/css/_abstract.scss
+++ b/css/_abstract.scss
@@ -8,3 +8,37 @@ $blk-col:#000000;
  justify-content: center;
  align-items: center;
 }
+// MEDIA QUERY MANAGER
+/*
+phone 0 - 600px;
+tab-port < 900px;
+tab-land < 1200px;
+normal < 1800px;
+big-desk > 1800px
+//default 1rem = 16px; 
+inside media query rem and em are same, set to browser defaule
+but research suggests em is better than rem in media query;
+*/
+@mixin respond ($break-point){
+    @if $break-point == phone {
+        @media (max-width: 37.5em) {//600px/16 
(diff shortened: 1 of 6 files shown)


rajgthub:
Thank you very much for the positive comments. Yes, I can fix it.

rajgthub:
I am unsure whether I made the pull request correctly! I think I have done my best to fix navigation section and the layout for the cases considering small devices. I have to find a way to make my navigation  collapse. Thanks for spending your valuable time on  checking my code and giving feedback to improve my knowledge further.

css/_abstract.scss
```diff
@@ -7,4 +7,38 @@ $blk-col:#000000;
...
+        @media (max-width: 75em) {//1200px/16
+            @content
+        };
+    }
+    @if $break-point == big-desk {
+        @media (min-width: 112.5em) {//1800px/16
+            @content
+        };
+    }
+
```
LucyMac: This seems like overkill Raj, I would try something simpler. If you want to store your breakpoints inside variables, that's good practice. For example:
```
$max-width-mobile: 767px;

@media (max-width: $max-width-mobile) {
    margin-left: 0;
  }
``` 
this is more semantic than `@include respond(phone) {`
With `@include respond(phone) {` I do not know if we are talking about a size smaller than a phone or bigger than a phone. If you use the term `max-` or `min-` it's easier to understand what's going on 🙂

css/style.css
```diff
@@ -432,18 +440,33 @@ th {
...
 
 .container {
-  width: 60%;
+  width: 65%;
   display: flex;
   justify-content: space-between;
 }
 
+@media (max-width: 56.25em) {
+  .container {
```
LucyMac: I find that using `em` for breakpoints makes it harder to understand. Consider using pixels which are more predictable and are widely used in media queries.

css/style.css
```diff
@@ -432,18 +440,33 @@ th {
...
   display: flex;
   justify-content: center;
   align-items: center;
 }
 
+@media (max-width: 37.5em) {
+  .navigation {
+    opacity: 0;
+    visibility: hidden;
+  }
```
LucyMac: I suppose you did this in anticipation of collapsing your menu? 🙂

css/_abstract.scss
LucyMac: I'm not sure the `@if` statements are necessary either. Keep it simple!

css/style.css
```diff
@@ -542,12 +593,31 @@ button:hover {
...
   list-style: none;
   display: flex;
   justify-content: space-between;
 }
 
+@media (max-width: 37.5em) {
+  .cases__devices {
+    flex-direction: column;
+  }
+}
```
LucyMac: Good, this looks great! 

Watch out for the spacing around that section - there is no padding at the top and around the headline, and a lot of whitespace at the bottom. Try and even the space out.

css/style.css
```diff
@@ -542,12 +593,31 @@ button:hover {
   align-self: center;
 }
 
+@media (max-width: 56.25em) {
+  .cases__content > h2 {
+    font-size: 2rem;
+    margin: 0;
+  }
```
LucyMac: Give this headline some breathing space 🌬

rajgthub: Yes, surely!

css/style.css
rajgthub: Yes, I will add padding.

css/style.css
rajgthub: Yes, that is right. I will fix it later.

css/style.css
rajgthub: I read something about using px and rem in media query in the link: https://zellwk.com/blog/media-query-units/
If you read the following on the page: (2 and 3)
2. User Zooms In
The original idea behind em based sizes was due to older browsers not being able to update pixel values when a user zooms. In this regard, testing the difference between media query units when a user zooms will help to answer the question on whether we can use px based media queries now.
3. User Changed Their Browser’s Font Setting

Concluding The Experiments
As you can see from our tests above, the only unit that performed consistently across all four browsers is em. There aren’t any differences between em and rem with the exception of bugs found on Safari.

px media queries performed well in two of the three experiments (with the exception of Safari, again). Unfortunately, px media queries remained at 400px in the third experiment, which makes it a no-go if you intend to support users who change their browser’s font-size value.

Hence, my conclusion after these experiments is: Use em media queries.

css/_abstract.scss
rajgthub: For this simple case, it is not necessary.

rajgthub:
I have added more changes to the code. Please have a look at them when you get time. Thanks very much!

PR #4, opened 2018-04-07
Francisca Vargas

franciscavargas:


css/style.css
```diff
@@ -1,14 +1,278 @@
...
 
-/*  TIP: When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex' */
+    .benefits ul {
+        flex-direction: column;
+    }
 
+    .benefits li {
+        padding: 2rem 0;
+    }
+}
```
LucyMac: The benefits look great on mobile! 
It would be good to have another breakpoint in between though, as they look a bit squished around 500px wide. Maybe a breakpoint at 600 or 768px (tablet size)?

css/style.css
```diff
@@ -1,14 +1,278 @@
...
+    align-items: center;
+    background-image: url(../img/first-background.jpg);
+    background-position: center bottom;
+    background-repeat: no-repeat;
+    background-size: cover;
+    color: white;
+    flex-direction: column;
+    height: 41.5rem;
+    justify-content: center;
+}
```
LucyMac: It would be nice to reduce the height of the hero on smaller screens.

css/style.css
```diff
@@ -1,14 +1,278 @@
...
+    border-bottom: 1px solid #DDD;
+    justify-content: center;
+    min-height: 70px;
+    position: fixed;
+    width: 100%;
+}
+
+nav .container {
+    justify-content: space-between;
+}
```
LucyMac: This looks good, though below 1200px it maybe causes the nav items to be a little too stretched away from each other? You could use `flex-end` and keep that alignment more consistent.

css/style.css
franciscavargas: Already done on the new version with flex

franciscavargas:
Hi Lucy! Not sure if the second PR was done correctly as it's appearing inside the first one? Thanks

PR #5, opened 2018-04-11
Karma than

than2042:
Hi Lucy, 
I made a new branch and change a few things, but my nav-bar is something not right. Could you please give me advice.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -121,7 +121,10 @@ body {
     border-radius: 100%;
     width: 1.5rem;
     height: 1.5rem;
-    display: inline-block;
+    display: inline-flex;
+    justify-content: center;
+    align-items: center;
+    align-content: center;
     margin: 1.25rem 5px;
     border: 1px solid  #248f8f;
     padding: 0.625rem 0.625rem 0.625rem 0.625rem;
@@ -138,7 +141,7 @@ body {
       /* margin-top: -4rem; */
     }
     .navigation .links {
-      display: flex
+      display: flex;
     }
     .intro h1 {
       text-align: left;
@@ -168,7 +171,7 @@ body {
       width: auto;
     }
     .cases .content {
-      margin-top: 5rem;
+      margin-top: 1rem;
     }
     .cases h2 {
       font-size: 19px;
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -2,14 +2,76 @@
...
+        <h1>Introducting Karma</h1>
+        <h2>Bring WiFi with you, everywhere you go.</h2>
+        <div class="button">
+            <button>Learn More</button>
+        </div>
+</section>
+</div>
+<section>
+    <div class="container">
+        <h2>Everyone needs a little Karma</h2>
```
LucyMac: the container class should wrap around your entire page. Move it up so it contains everything else (after `body`)

index.html
```diff
@@ -2,14 +2,76 @@
...
+                <h3>Pay as You Go</h3>
+            </div>
+        </div>
+    </div>
+</section>
+<footer class="social">
+    <div class="content">
+        <hr>
+        <h5>Join us on</h5>
+        <ul>
```
LucyMac: remember `<ul>`s have a default `padding-left`. Set it to 0 and your icons will be centered.

css/style.css
```diff
@@ -1,11 +1,159 @@
...
+}
+
+.main {
+  background-image: url('../img/first-background.jpg');
+  background-position: center bottom;
+  background-repeat: no-repeat;
+  background-size: cover;
+  height: 50.35rem;
+  width: auto;
+  padding: 20rem 5rem 5rem 3rem;
```
LucyMac: You don't need this! Remove width, padding and margin and use your `col-` classes instead.

css/style.css
```diff
@@ -1,11 +1,159 @@
...
+  margin: 0 auto;
+  
+}
+
+.nav-bar .links li {
+  display: inline-flex;
+  color: #504d4b;
+  font-size: 1.5rem;
+  margin-left: 1.3rem;
+}
```
LucyMac: Your nav is on the right track - you just need to align the `ul` to the right, and vertically centered (and the `li`s maybe a little smaller) and it should be fine!

PR #20, opened 2020-01-26
first change

leealex88:
kfsbaklfbsklf


css/style.css
```diff
@@ -1,11 +1,261 @@
...
+    position: absolute;
+    transition: all .2s ease-out;
+    top: -5px;
+}
+
+.header .menu-btn {
+    display: none;
+}
+
+.header .menu-btn:checked~.menu {
```
LucyMac: Nice - this is pretty advanced CSS!

css/style.css
```diff
@@ -1,11 +1,261 @@
...
+}
+
+@media (max-width: 650px) {
+    .flex-container {
+        flex-direction: column;
+    }
+}
+
+section div div {
+    margin: 20px;
```
LucyMac: Watch out for rules that are applied to generic elements such as `div` or `section` - you may not _always_ want the margin on all divs for example.

index.html
```diff
@@ -1,21 +1,101 @@
...
+          width="26"
```
LucyMac: Avoid having this amount of image code in your markup - could you have simply linked the `src` attribute of your img element to '/img/karma-logo.svg'?

PR #21, opened 2020-01-29
Karma, what's doing around, coming around

yusuf963:
here is(I believe) the copy of "Karma" landing page,  I was missing a step in pushing part and took a good while to figure it out

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,10 +1,114 @@
 
  /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
 
-body {
+ body {
     font-family: 'Roboto', sans-serif;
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -1,10 +1,114 @@
...
+   font-size: 45px;
+   font-family: 'Lucida Sans', 'Lucida Sans Regular', 'Lucida Grande', 'Lucida Sans Unicode', Geneva, Verdana, sans-serif;
+   font-weight: 500;
+}
+h4{
+    font-size: 16px;
+    font-weight: 400;
+    margin-left: 50px;
+}
+spam{
```
LucyMac: typo: spam --> span

PR #24, opened 2020-01-29
Karma website part 1

Alisina92:
I am currently developing on it lucy sorry is incomplete but will complate it


index.html
```diff
@@ -10,12 +15,35 @@
...
+      <li class="nav-list">Store</li> 
+      <li class="nav-list">Blog</li>
+      <li class="nav-list">Help</li>
+      <li class="nav-list">Login</li>
+    </ul class="menu-nav">
+  </nav>
+  <main class= web-content>
+    <section class = content-container>
+      <img class='background-img' src="img/first-background.jpg" alt= 'karma-wifi-device'>
+          <h1><div class =first-list-element>Introducing Karma</h1></div>
```
LucyMac: You're missing the quotation marks here

index.html
```diff
@@ -1,6 +1,11 @@
+<!-- Add your HTML markup here -->
+<!-- Remember: Use semantic HTML tags like <header>, <main>, <nav>, <footer>, <section> etc -->
+<!-- All the images you need are in the 'img' folder -->
+
+
 <!DOCTYPE html>
 <html lang="en">
-<head>
+<header>
```
LucyMac: Be careful not to confuse `header` and `head`.

`head` is for the meta information and linking fonts / stylesheets etc.

`header` is part of your page and should include your nav etc.

index.html
```diff
@@ -31,14 +31,23 @@
   </nav>
   <main class= web-content>
     <section class = content-container>
```
ChrisOwen101: Make sure you don't have spaces here. It should look like this:
    <section class=content-container>

index.html
Alisina92: Thanks for the feedback I will implement it accordingly.

index.html
Alisina92: Thanks for the feedback. I have fixed these but still my :
<section>
        <ul>
          <li><h1 class="karma-title">Everyone needs a little Karma</h1></li>
          <li></li>
          <li></li>
          <li></li>
          <li></li>
          <li></li>
        </ul>
     
    </section>
is hidden under the main background.

LucyMac: Your classes all need quotation marks around them @Alisina92 
`<main class="web-content">`
`<section class="content-container">`

You have typing errors on most lines. As Chris said:
- You should never have spaces when you are writing HTML attributes (for example adding class, or src, or href). The format should always be attributename="attributevalue"
e.g. `class="main-content"`
- You need the quotation marks

css/style.css
```diff
@@ -16,4 +8,135 @@ body {
...
+.web-content {
+  overflow-x: hidden;
+  overflow-y: scroll;
+  height: 100%;
+}
+
+.background-img {
+  width: 100%;
+  position: absolute;
+}
```
LucyMac: Instead of using an `img` tag for the large picture, try setting it as a `background-image` instead.

css/style.css
```diff
@@ -16,4 +8,135 @@ body {
...
+  width: 7%;
+}
+
+.logo-icon {
+  margin-right: 200px;
+}
+
+.web-content {
+  overflow-x: hidden;
+  overflow-y: scroll;
```
LucyMac: Do you need these overflow rules? Once you fix your hero section you will find you don't need this.

css/style.css
LucyMac: It shouldn't need absolute positioning. It's better to let the page flow naturally.

css/style.css
```diff
@@ -16,4 +8,135 @@ body {
...
+
+.first-list-element {
+  position: absolute;
+  z-index: 100;
+  left: 550px;
+  top: 350px;
+  color: white;
+}
+* .second-list-element {
+  position: absolute;
```
LucyMac: What are these elements with `position: absolute`?

Use flexbox for positioning instead!

css/style.css
```diff
@@ -16,4 +8,135 @@ body {
...
+}
 
+.second-content-container {
+  position: relative;
+  z-index: -1;
+  margin-top: 60%;
+}
+
+.content-container .btn {
+  position: absolute;
```
LucyMac: Too much absolute positioning Alisina - try other options first! 

In many cases, Flexbox will be the solution.

css/style.css
```diff
@@ -16,4 +8,135 @@ body {
...
+  display: flex;
+  justify-content: center;
+  max-width: 100%;
+}
+
+.nav-list {
+  padding: 10px;
+  padding-left: 10px;
+  padding-right: 1px;
+  width: 7%;
```
LucyMac: You don't need to set a width for each element. They should each grow to the size of their text content.

css/style.css
```diff
@@ -16,4 +8,135 @@ body {
...
+  font-size: 16px;
+}
+
+.nav-flex-box {
+  display: flex;
+  justify-content: center;
+  max-width: 100%;
+}
+
+.nav-list {
```
LucyMac: Choose your naming carefully - you are using this class on each list _item_, not on the list itself. So it would be more appropriate to name this class `nav-list-item`.

index.html
```diff
@@ -1,21 +1,93 @@
...
+            </figure>
+            <figure>
+              <li><img class ="footer-list" src="img/facebook.svg"></li>
+            </figure>
+            <figure>
+              <li><img class ="footer-list" src="img/instagram.svg"></li>
+            </figure>
+        </ul>
+        <h5>© Karma Mobility, Inc.</h5>
+        </footer>
```
LucyMac: Good structure with good use of header / nav / main / footer 👍

PR #25, opened 2020-01-29
Class 6/rayhan

Rayhan94:


css/style.css
```diff
@@ -17,3 +17,119 @@ body {
...
+    font-size: 1rem;
+    margin-top: 25px; 
+    -webkit-font-smoothing: antialiased;
+  }
+  
+  .main-navigation li {
+    display: inline-block;
+    margin-left: 35px;
+  }
+
```
amcraethanan: Try re-sizing your page to a smaller size, does the navigation menu items move so that they are still visible? Have a think about how you could use a flexbox container to help the menu options flow with your page in your navigation class. 

Have look at adding some side padding as well so that there space around the elements when looking at smaller screens.

css/style.css
```diff
@@ -17,3 +17,119 @@ body {
...
+  
+  .devices > li {
+    width: 33%;
+    box-sizing: border-box;
+    height: 4em;
+    display: inline-block;
+    list-style: none;
+    box-sizing: border-box;
+    margin-left:-5%;
+    margin-right: -5%;
```
amcraethanan: `box-sizing` appears twice in this class, you only need one instance to have this style applied.

css/style.css
```diff
@@ -17,3 +17,119 @@ body {
...
+  }
+  .socialmedia .icons {
+    text-align: center;
+  }
+  .socialmedia .icons > li {
+    border-radius: 1.25rem;
+    width: 2.5rem;
+    height: 2.5rem;
+    display: inline-block;
+    margin: 1.25rem 5px;
```
amcraethanan: You shouldn't mix the units used here, have a read about the [differences](https://medium.com/code-better/css-units-for-font-size-px-em-rem-79f7e592bb97) between rem and px. In this case you probably want to use `rem` but using something like flexbox could also help you out here to get evenly-spaced items.

PR #26, opened 2020-01-30
Kyaw karma

kiowd:
My First pull request

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,12 +1,22 @@
 
  /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
 
+ 
 body {
     font-family: 'Roboto', sans-serif;
     -webkit-font-smoothing: antialiased;
 }
+/**
+ * Add your custom styles below
+ * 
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -16,4 +16,197 @@ body {
...
+
+#linkOne {
+    margin-left: 14rem;
+    color:black;
+    
+}
+#linkOne,
+#linkTwo, #linkThree, #linkFour, #linkFive, #linkSix {
+    margin: 0px 15px;
+}
```
amcraethanan: instead of targeting the elements by id, as all these elements are meant to the same anyway, consider creating a class to capture this styling instead.

css/style.css
```diff
@@ -16,4 +16,197 @@ body {
...
+    
+}
+
+a:hover {
+    color: darkslategray
+    
+}
+
+.navContainer {
+    display: flex;
```
amcraethanan: well done on using container with flexbox 👍

PR #29, opened 2020-01-31
Class 6/tam

TamBahri:
Create Header, logo and nav step

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,8 +1,7 @@
 
- 
  /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
 
-body {
+ body {
     font-family: 'Roboto', sans-serif;
     -webkit-font-smoothing: antialiased;
 }
@@ -33,7 +32,7 @@ header {
 
 .logo > img {
   width: 100%;
-  height: 4.375rem;
+  height: 3.375rem;
   max-width: 200px;
   display: flex;
   justify-content: flex-start;
@@ -44,11 +43,13 @@ header {
 
 .nav-bar{
   background-color: #ffffff;
-  height: 4.375rem;
-  width: 60rem;
+  height: 3.375rem;
+  width: 70rem;
 	padding: 0 0.938rem;
 	margin: 0 auto;
-	font-family: 'Roboto', sans-serif;
+    font-family: 'Roboto', sans-serif;
+    font-weight: 500;
+  
   
   
 }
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -12,10 +12,57 @@
...
+    <header>
+        <div class="logo">
+          
+          <img class="logo" src="http://swarley00.github.io/startup-landing-page-clone/images/karma-logo.svg" alt="Logo Image">
+            
+        </div>
+        <nav class = "nav-bar">
+            <ul>
+                <li>Meet Karma</li>  <li>How-it-Works</li> 
+ <li>Store</li>
```
LucyMac: Be careful, indentation is really important to be able to read code and spot mistakes easily

index.html
```diff
@@ -12,10 +12,57 @@
...
 
+        </nav>
+      
+    </header>
+  <section class="introduction">
+    <div class= "intro-content">
+        <h1>Introducing Karma</h1>
+        <h3>Bring WiFi with you, everywhere you go.</h3>
+
+        <div class="btn">
```
LucyMac: Use a `<button>` element here instead of a div - and you don't need the `<p>` inside.

css/style.css
```diff
@@ -1,10 +1,142 @@
...
+  justify-content: flex-start;
+  align-items: center;
+  text-align: center;
+  margin-left: 20px;
+}
+
+.nav-bar{
+  background-color: #ffffff;
+  height: 3.375rem;
+  width: 70rem;
```
LucyMac: Your nav-bar is too wide here, it's breaking out of the screen. This is because of a couple of things:

- Your `<header>` needs a width of 100% so it's always the same width as the screen.
- The `nav-bar` section is fixed at 70rem and doesn't take into account the fact that the screen may be smaller than 70rem (then the nav items get cut off) or larger.

A good fix could be to remove the width on the nav-bar, and set the header to `width: 100%`. Once you've done that, can you think which flexbox rule you could use to ensure the logo always stays on the left, and the nav-bar always stay on the right?

css/style.css
```diff
@@ -1,10 +1,142 @@
...
+}
+
+li:hover {
+	cursor: pointer;
+	color: #333333;
+}
+
+
+.introduction  {
+    background-image: url("http://swarley00.github.io/startup-landing-page-clone/images/first-background.jpg");
```
LucyMac: Why not link to the image in your img folder?

LucyMac: It works in the same way: `url('./img/image-name.jpg')`

Note: This is just an example and may not be the exact path.

css/style.css
```diff
@@ -1,10 +1,142 @@
...
+    background-size: cover;
+    background-repeat: no-repeat;
+    height: 582.4px;
```
LucyMac: Avoid using decimal places with pixels.

css/style.css
```diff
@@ -1,10 +1,142 @@
...
+}
+@media (max-width: 500px) {
+  header {
+    flex-wrap: wrap;
+  }
+}
+
+  
+.logo {
+    margin-top: 15px;
```
LucyMac: This rule is forcing the logo down so it is not centered vertically in the nav bar.

css/style.css
```diff
@@ -1,10 +1,142 @@
...
+  }
+
+
+.logo > img {
+  width: 100%;
+  height: 3.375rem;
+  max-width: 200px;
+  display: flex;
+  justify-content: flex-start;
+  align-items: center;
```
LucyMac: These rules only work when applied to a parent element, which wraps around child elements.
The image is not a parent so these rules have no effect.

Remember, when you want to move elements with flexbox, apply the flexbox rules to the **parent** element, not the element you want to move!

css/style.css
TamBahri: many times told me path did not exist.

PR #30, opened 2020-01-31
Scavus_karma

SCavus:
This is the latest version. There might be unnecessary CSS commands as well as divs.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -39,13 +39,6 @@ article, aside, details, figcaption, figure, footer, header, hgroup, menu, nav, 
     border: 0;
     width: 26;
 }
-/*.content {
-    padding: 0 1;
-    margin: 1120;
-    color: lightslategrey);
-    display: flex;
-    flex-direction: row;   
-}*/
 
 li{
     list-style-type: none;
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -12,10 +12,56 @@
     <link rel="shortcut icon" type="image/x-icon"  href="favicon.ico">
 </head>
 <body>
-
+    <nav class="navbar">
```
amcraethanan: Along the lines of thinking of semantic html, the menu options are always at the top of the page in the Karma website - logically the `nav` element could very well fit within a `<header>` element.

css/style.css
```diff
@@ -16,4 +16,137 @@ body {
...
+    flex-direction: row;
+    justify-content: center;
+    margin: 0 300px 200px 300px;
+}
+
+.captions {
+    background-color: white;
+    display: inline-block;
+    flex-direction: column;
+    justify-content: space-around;
```
amcraethanan: as display has been set to `inline-block` so these two flex properties won't quite work

index.html
```diff
@@ -12,10 +12,56 @@
...
+        <div>
+        <img class="logo" src="img/karma-logo.svg" alt="logo image">
+        </div>
+        <div class="content">
+            <ul class="links"></ul>
+                <li>Meet Karma</li>
```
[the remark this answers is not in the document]
SCavus: </ul> was an automated tag. There is also one after the <li>s. I have updated accordingly.

PR #31, opened 2020-02-01
Flexbox/karma/Mo

mnashawati:
Homework is done.. I am still working on responsiveness  and media queries, but would really appreciate code review and some feedback. Thank you in advance! :)


css/style.css
```diff
@@ -16,4 +10,209 @@ body {
...
+ul li {
+  padding: 1rem;
+}
+
+.bold {
+  font-weight: bolder;
+}
+
+/* First section */
+#hero {
```
LucyMac: Use classes instead of IDs

css/style.css
```diff
@@ -16,4 +10,209 @@ body {
...
+  background-image: url("../img/first-background.jpg");
+  background-size: auto;
+  background-position: center;
+  height: 112vh;
```
LucyMac: Why 112vh? 100vh would work well :)

Also watch out, you'll need `background-repeat: no-repeat` for the image to show only once.

css/style.css
```diff
@@ -16,4 +10,209 @@ body {
...
+
+  *,
+  html,
+  body {
+    font-size: 12px;
+  }
+}
+
+@media (max-width: 350px) {
+
```
LucyMac: Great that you added media queries :)

index.html
```diff
@@ -1,21 +1,78 @@
...
+      <img class="logo" src="img/karma-logo.svg" alt="karma-logo" />
+      <ul>
+        <li class="bold"><a href="">Meet Karma</a></li>
+        <li><a href="">How it Works</a></li>
+        <li><a href="">Store</a></li>
+        <li><a href="">Blog</a></li>
+        <li><a href="">Help</a></li>
+        <li><a href="">Login</a></li>
+      </ul>
+    </nav>
```
LucyMac: Nice nav structure

index.html
```diff
@@ -1,21 +1,78 @@
...
+  </header>
 
-    <!-- Add your HTML markup here -->
-    <!-- Remember: Use semantic HTML tags like <header>, <main>, <nav>, <footer>, <section> etc -->
-    <!-- All the images you need are in the 'img' folder -->
+  <main>
+    <section id="hero">
```
LucyMac: Don't use IDs! Always choose classes over IDs for CSS.

index.html
```diff
@@ -1,21 +1,78 @@
...
+      <h1>Introducing Karma</h1>
+      <h3>Bring WiFi with you, everywhere you go.</h3>
+      <button type="button" class="button"> <a href="#">Learn More</a>
```
LucyMac: Choose between a button or a link here, not both. 
The general rule of thumb is: 
- If clicking it takes you to a different page, use an anchor tag `<a>`
- If clicking it makes something happen like open a modal, or toggle something on the page (basically a Javascript action) then use a button.

You can however use an anchor tag and give it a class of `button` and in your CSS define `a.button {}`

index.html
```diff
@@ -1,21 +1,78 @@
...
+    </section>
+    <section id="benefits">
+      <h2>Everyone needs a little Karma.</h2>
+      <ul class="icons">
+        <li>
+          <figure>
+            <img class="icon" src="img/icon-devices.svg" alt="icondevices" />
+            <figcaption>Internet for all devices</figcaption>
```
LucyMac: I'm not sure `figcaption` is the right choice here. Is the text really describing what the image is showing? Or is the image illustrating what the text is saying?

index.html
```diff
@@ -1,21 +1,78 @@
...
+            <img class="icon" src="img/icon-refill.svg" alt="icon-refill" />
+            <figcaption>Pay as You Go</figcaption>
+          </figure>
+        </li>
+      </ul>
+    </section>
+  </main>
 
+  <footer>
+    <div>
```
LucyMac: Good structure for the footer, except for this `div` - what does it do? Is it necessary?

mnashawati:
Thank you so much for the feedback @LucyMac I really appreciate it :D

css/style.css
mnashawati: 100vh is not displaying as much of the image as in your original page for some reason..

PR #33, opened 2020-02-01
Class6/Abdullah

abcelen:
I couldnt adjust nav link, space in main area.


index.html
```diff
@@ -12,10 +12,71 @@
...
+        <div class="container">
+            <h1>Introducing Karma</h1>
+            <h2>Bring WiFi with you, everywhere you go.</h2>
+            <button class="hero__button">Learn More</button>
+        </div>
+    </div>
+    <!-- Main starts here-->
+    <div class="main">
+        <div class="container">
+            <h1 class="main__title">Everyone needs a little Karma.</h1>
```
LucyMac: This heading is slightly smaller than the hero heading, so it could be an h2

LucyMac: Also, be careful - you added a very large top padding on all h1s in your CSS, so this heading is pushed down too low on the page.
How could you apply the top padding to the hero h1 only, and not this one?

index.html
```diff
@@ -12,10 +12,71 @@
...
+    <header class="header"> <!-- contains logo and nav -->
+        <div class="container">   
+            <div class="header__inner">
+               <img src="img/karma-logo.svg" width="25" alt="karma-logo">
 
-    <!-- Add your HTML markup here -->
-    <!-- Remember: Use semantic HTML tags like <header>, <main>, <nav>, <footer>, <section> etc -->
-    <!-- All the images you need are in the 'img' folder -->
+                <nav class="nav">
+                    <a  href="#">Meet Karma</a>
```
LucyMac: Good use of the nav tag. Could you improve this to include an unordered list and list items?

index.html
```diff
@@ -12,10 +12,71 @@
...
+                    <a  href="#">Store</a>
+                    <a  href="#">Blog</a>
+                    <a  href="#">Help</a>
+                    <a  href="#">Login</a>
+                </nav>
+            </div>
+        </div>
+    <!-- header starts here-->
+    </header>
+        <div class="hero">
```
LucyMac: Why is the hero not inside `main`?
Anything that isn't the header or the footer can go inside main.

index.html
```diff
@@ -12,10 +12,71 @@
     <link rel="shortcut icon" type="image/x-icon"  href="favicon.ico">
 </head>
 <body>
+    <header class="header"> <!-- contains logo and nav -->
+        <div class="container">
```
LucyMac: Great use of the container class in each section 👍

css/style.css
```diff
@@ -6,12 +6,197 @@ body {
...
+    margin:  auto;
+    display: inline-flex;
+    justify-content: space-evenly;
+    align-items: center;
+    
+
+}
+
+.main__block{
+    width: 20vw;
```
LucyMac: Do you need a fixed width on these elements? Maybe some side padding would be enough and allow more flexibility.

css/style.css
```diff
@@ -6,12 +6,197 @@ body {
...
+    display: flex;
+    flex-direction: column;
+
```
LucyMac: What could you add here to make sure the icon is centered horizontally each block?

css/style.css
```diff
@@ -6,12 +6,197 @@ body {
...
+    display: flex;
+    flex-direction: column;
+    
+}
+
+.main__text{
+    font-weight: 400;
+    font-size: 1.563rem;
+    color: #1d1e21;
+    width: 250w;
```
LucyMac: the unit `w` doesn't exist. 
Do you need a width here?

index.html
```diff
@@ -12,10 +12,71 @@
...
+                </div>
+            </div>
+        </div>
+    </div>
+</div>
+ <!-- footer starts here-->   
+<footer class="footer">
+    <div class="container">
+        <div class="footer__title">
+            Join us on
```
LucyMac: could this go inside a more semantic text tag?

css/style.css
```diff
@@ -6,12 +6,197 @@ body {
...
+.nav {
+    font-size: 16px;
+
+}
+
+.nav a {
+    color: #5e5c5c;
+    display: inline-block;
+    vertical-align: top;
+    margin: 0 10px;
```
LucyMac: can you add a rule to remove the underline on the links?

index.html
```diff
@@ -12,10 +12,71 @@
...
+    </header>
+        <div class="hero">
+        <div class="container">
+            <h1>Introducing Karma</h1>
+            <h2>Bring WiFi with you, everywhere you go.</h2>
+            <button class="hero__button">Learn More</button>
+        </div>
+    </div>
+    <!-- Main starts here-->
+    <div class="main">
```
LucyMac: Use the HTML tag `main` instead of a div here.

css/style.css
```diff
@@ -6,14 +6,202 @@ body {
...
+    margin: 1.625rem 0 3.125rem;
+    font-weight: 300;
+    text-align: center;
+    color: #1d1e21;
+    margin-top: 120px;
+}
+
+
+.main__inner{
+    width: 900px;
```
LucyMac: This rule is breaking the page because it applies for all breakpoints - even when the screen is smaller than 900px!

css/style.css
```diff
@@ -194,11 +194,14 @@ h2 {
...
- * Add your custom styles below
-
- * 
- * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
- */
 
+@media (max-width: 768px) {
+    .main__inner{
+    display: flex;
+    flex-direction: column;
```
LucyMac: Good addition. Can you fix the width issue which is causing the page to break?

PR #35, opened 2020-02-01
commit

jreo123:


index.html
```diff
@@ -3,19 +3,60 @@
...
+        <img src="favicon.ico" alt="Karma logo">
+        <ul class="navbut">
+        
+            <li>Meet Karma</li>
+            <li>How it Works </li>
+            <li>Store</li>
+            <li>Blog</li>
+            <li>Help</li>
+            <li>Login</li>
+
```
LucyMac: Nice use of an unordered list. Can you add links inside each item, and give them a hover effect?

index.html
```diff
@@ -3,19 +3,60 @@
...
     <link rel="stylesheet" href="css/style.css">
     <link rel="shortcut icon" type="image/x-icon"  href="favicon.ico">
-</head>
+</head> 
+
+<header>
+</header>
+
+
+<nav>
```
LucyMac: The nav needs to be inside the body tag.
The `head` tag wraps all meta information (not displayed on the page, just tells the browser about the page). Everything you want to _appear_ on the page should live inside the `body` tag

index.html
```diff
@@ -3,19 +3,60 @@
...
+            <li>How it Works </li>
+            <li>Store</li>
+            <li>Blog</li>
+            <li>Help</li>
+            <li>Login</li>
+
+        </ul>
+    </nav>
 <body>
+
```
LucyMac: Wrap the next 2 sections inside the `main` tag.

Anything in the body that is not the header or the footer can go inside the main tag

index.html
```diff
@@ -3,19 +3,60 @@
...
+        <h1>Introducing Karma</h1>
+        <h2>Bring Wifi with you, everywhere you go.</h2>     
+    </div>
+
+</section>
 
+<section>
+    <h1>Everyone needs a little Karma.</h1>
+        <ul class="imgpic">
+            <li><img src="/img/icon-devices.svg" alt="internet devices"></li>
```
LucyMac: You need to change something in the image path here for something to appear on the page. 
You'll also need to add a width to these images in CSS, so the browser knows how big to make them - otherwise they may not appear at all as their width will be `0`.

PR #36, opened 2020-02-01
Flexbox/karma/abbas mokhtari

abbasmokhtari:
I have made minor changes and since some logos were not available I downloaded those logos and they are different to the page sent as guide


css/style.css
```diff
@@ -6,6 +6,153 @@ body {
...
+    text-align: center;
+}
+
+.dev {
+    width: 70%;
+}
+
+.coffee {
+width: 70%;
+}
```
LucyMac: Why define 2 separate classes that do the same thing here? You can create one class (for example `segment-icon`) and set the size once.
Remember, CSS classes are meant to be re-used.

index.html
```diff
@@ -1,21 +1,112 @@
...
+                </li>
+                <li>
+                    <a href="#">Help</a>
+                </li>
+                <li>
+                    <a href="#">Login</a>
+                </li>
+            </ul>
+        </nav>
+    </header>
```
LucyMac: Very well structured header and nav

index.html
```diff
@@ -1,21 +1,112 @@
...
+
+    <!-- SHOWCASE -->
```
LucyMac: Use `main` for anything that is not the header or the footer

index.html
```diff
@@ -1,21 +1,112 @@
...
+        </h4>
+
+        <button class="button" type="button">
+            Learn More
+        </button>
 
+    </section>
+
+    <!-- SEGMENTS -->
+    <section class="seg">
```
LucyMac: Be careful with shortened class names - I wasn't sure what 'seg' meant when I first read it.

index.html
```diff
@@ -1,21 +1,112 @@
...
+                    Internet for all devices
+                </h5>
+            </div>
+            <div class="cup">
+                <img src="img/icon-coffee.svg" alt="Coffee" class="coffee">
+                <h5>
+                    Boost your productivity
+                </h5>
+            </div>
+            <div class="fuel">
```
LucyMac: Do you need 3 different classes for `devices`, `cup` and `fuel`?

index.html
```diff
@@ -1,21 +1,112 @@
...
+                <img src="img/icon-refill.svg" alt="refill" class="nuzzle">
+                <h5>
+                    Pay as You Go
+                </h5>
+            </div>
+        </div>
+    </section>
+
+    <!-- FOOTER -->
+    <footer id="footer">
```
LucyMac: Why are you using an ID here? 

Don't use IDs in your CSS!
https://stackoverflow.com/questions/8279132/why-shouldnt-i-use-id-selectors-in-css

index.html
```diff
@@ -1,21 +1,112 @@
...
+                <img src="img/facebook-circled.png" alt="Facebook logo" class="face">
+            </a>
+            <a href="#">
+                <img src="img/Insta-logo.png" alt="Instagram" class="insta">
+            </a>
+        </div>
+        <p class="copy">
+            &copy; Karma Mobility, Inc.
+        </p>
+        <div class="end"></div>
```
LucyMac: Why add an empty div here?
If you want to add spacing, add it to the footer.
example: 
```
footer {
   padding-bottom: 100px;
}
```

PR #37, opened 2020-02-01
Developing

MahmutSel404:
branch merging


css/style.css
```diff
@@ -16,4 +12,102 @@ body {
...
+
+
+/**
+the button
+*/
+.buttons{
+    color:white;;
+    text-align:center;
+    margin-top:1.563rem;
+}
```
LucyMac: Here you could have styled the `button` element directly, instead of the parent element `.buttons`.
Can you add a `background-color` to the button so it looks like the original?

index.html
```diff
@@ -12,7 +12,71 @@
...
 <body>
+    <nav class="top-navigation">
+        <div class="content">
+            <ul class="links">
+                <li class="active">Meet Karma</li>
+                <li>How it Works</li>
+                <li>Store</li>
+                <li>Blog</li>
+                <li>Help</li>
+                <li>Login</li>
```
LucyMac: Can you make these links as well, and give them a hover style?

index.html
```diff
@@ -12,7 +12,71 @@
...
+    <section class="section-1">
+        <div class="content">
+            <h1>Introducing Karma</h1>
+            <h2>Bring WiFi with you, everywhere you go.</h2>
+            <div class="buttons">
+                <button>Learn More</button>
+            </div>
+        </div>
+    </section>
+    <section class="section-2">
```
LucyMac: Avoid naming elements by order of their appearance on the page, as this may change when you switch from mobile view to desktop view. It also doesn't give much indication of what is inside the section - try and be a little more descriptive.

index.html
```diff
@@ -12,7 +12,71 @@
...
+                <li>
+                    <img width="100" height="100" src="img/icon-coffee.svg" alt="Boost your productivity image">
+                    <div class="caption">
+                        <h5>Boost your productivity</h5>
+                    </div>
+                </li>
+                <li>
+                    <img width="100" height="100" src="img/icon-refill.svg" alt="Pay as You Go image">
+                    <div class="caption">
+                        <h6>Pay as You Go</h6>
```
LucyMac: Why are these 3 headlines all using different HTML tags, when they all look the same in the original design? 
Try to keep things consistent.

css/style.css
```diff
@@ -16,4 +12,102 @@ body {
...
+    color:black;
+    padding:none;
+    text-align:center;
+}
+.section-2 h3{
+    font-size:60px;
+    font-weight:normal;
+
+}
+
```
LucyMac: Could you add something here to make the 3 'devices' all lign up in a horizontal row?

index.html
```diff
@@ -12,7 +12,71 @@
...
+    </section>
 
+    <footer class="social-media">
+        <div class="content">
+            <hr>
+            <h5>Join us on</h5>
+            <ul class="icons">
+                <li class="twitter">
+                </li>
+                <li class="facebook">
```
LucyMac: You could make these links too.

PR #38, opened 2020-02-01
Class6/yohannes

byohannes:
I have done a simple website for a youth educational charity using semantic HTML.


css/style.css
```diff
@@ -1,19 +1,73 @@
...
- * Add your custom styles below
- * 
- * Remember: 
- * - Be organised, use comments and separate your styles into meaningful chunks
- *    for example: General styles, Navigation styles, Hero styles, Footer etc.
- * 
- * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
- */
+.nav .nav-links {
+  float: right;
```
LucyMac: Please don't use floats, they cause layout issues and have been replaced by flexbox which is much easier and more predictable to use. 
Can you edit this to use flexbox instead?

index.html
```diff
@@ -1,21 +1,79 @@
...
+          <li>Who WE Are</li>
+          <li>What We Do</li>
+          <li>Our Movement</li>
+          <li>Get Involved</li>
+        </ul>
+        <img class="e-logo" src="img/logo.png" alt="Logo Image" width="" />
+      </div>
+    </nav>
+    <section class="section-one">
+      <div class="sec1"></div>
```
LucyMac: You're duplicating markup unnecessarily here. `sec1` doesn't do anything - you can remove it.

index.html
```diff
@@ -1,21 +1,79 @@
...
+    </section>
+    <section class="section-two">
+      <div class="sec2">
```
LucyMac: Again, why 2 elements here, when only one would be enough?

css/style.css
```diff
@@ -1,19 +1,73 @@
...
+  margin-top: 7.875rem;
+  padding: none;
+  color: #1d1e21;
+  text-align: center;
+}
+.section-two .buttons {
+  text-align: center;
+  margin-top: 1.563rem;
+}
+.lowe {
```
LucyMac: Make sure you name things using easy-to-understand words. I'm not sure what `lowe` means.

css/style.css
```diff
@@ -1,19 +1,73 @@
...
 
+.section-two .sec2 {
+  margin-top: 7.875rem;
+  padding: none;
+  color: #1d1e21;
+  text-align: center;
+}
+.section-two .buttons {
+  text-align: center;
+  margin-top: 1.563rem;
```
LucyMac: Can you add styles to your button so it looks nicer?

byohannes:
Thanks, For the feedback Lucy. I am working to fix on all the mistakes and also to include media queries so the updated version of the homework will be up as a PR.

PR #39, opened 2020-02-01
Class 6/sabia

sabia-m:


css/style.css
```diff
@@ -6,6 +6,118 @@ body {
...
+}
+
+.banner {
+    position: relative;
+    text-align: center;
+    color: white;
+}
+
+.introduction {
+    position: absolute;
```
LucyMac: You shouldn't have to use absolute positioning here. 
If you switch to setting the large image as a `background-image` in CSS instead of using an `img` tag, then you'll be able to superimpose the text on top of it much more easily. 

Once you've done the image switch, change the positioning for `introduction` and `intro-subheading` so it uses flexbox instead.

css/style.css
```diff
@@ -6,6 +6,118 @@ body {
...
+}
+
+.join {
+    text-align: center;
+}
+
+
+.media {
+    text-align: center;
+}
```
LucyMac: These 2 classes do the same thing. You should create one class only, and re-use it several times.

index.html
```diff
@@ -10,8 +10,52 @@
...
+<header>
+    <img class="logo" src="img/karma-logo.svg" alt="karma logo">
+    <nav>
+        <ul>
+            <li>Meet Karma</li>
+            <li>How It Works</li>
+            <li>Store</li>
+            <li>Blog</li>
+            <li>Help</li>
+            <li>Login</li>
```
LucyMac: Can you add links inside the `li`s and give the nav items a hover effect?

index.html
```diff
@@ -10,8 +10,52 @@
...
+        </ul>
+    </nav>
+</header>
+
```
LucyMac: Remember to use the `main` tag to wrap anything that is not the header or the footer.

index.html
```diff
@@ -10,8 +10,52 @@
...
+<section class="banner">
+    <img class="first-banner" src="img/first-background.jpg" alt="first banner">
```
LucyMac: As mentioned above, remove this img tag and set the image to be the `background-image` of the `.banner` in your CSS instead. 
This will help you re-position the h1 and h2 which are currently off-center.

css/style.css
```diff
@@ -6,6 +6,118 @@ body {
...
+
+}
+
+
+.details {
+    padding: 10px;
+    margin: auto;
+    text-align: center;
+    font-size: 14px;
+    justify-content: stretch;
```
LucyMac: This rule won't do anything unless you use `display: flex` at the same time.

PR #44, opened 2020-02-02
karma_ellie

ellietms:


index.html
```diff
@@ -4,18 +4,83 @@
...
+            <button>Learn More</button>
+        </div>
+    </section>
+    <!-- Section two -->
+    <section class="cases">
+        <h1 class="cases_h1">Every one needs a little Karma</h1>
+        <ul>
+            <div class="devices">
+                <li>
+                    <img src="../icon-devices.svg" width="60">
```
ChrisOwen101: To make these images load properly you should locate them like this

```
  <img src="img/icon-devices.svg" width="60">
```

I see in the commit history that you spent a while trying to fix this. Don't forget to help to ask for help on Slack when you have trouble! 

It's better than we help with the small stuff like this so you can focus on the bigger problems.

ChrisOwen101:
This is a really awesome homework Ellie! You're totally moving in the right direction.

HTML/CSS can take a while to master so keep at it. Try fixing the links on your images and you'll see how much closer you feel.

css/style.css
```diff
@@ -1,19 +1,140 @@
...
+  .logo-image{
+    display: block; 
+    height: 20px;
+    width: 50px;
+    margin-inline-start: 300px;}
+
+
+    .introduction{
+        display: block;
+        background-image: url(img/first-background.jpg);
```
ChrisOwen101: You can fix this like this:

```
    background: url('../img/first-background.jpg') 
```

Relative links can be a little tricky. You can learn about them here:
https://css-tricks.com/quick-reminder-about-file-paths/

css/style.css
```diff
@@ -1,19 +1,140 @@
...
+        color: rgb(240, 232, 232);
+        text-align: center;
+        box-sizing: border-box;
+        padding: 0 0;
+        margin: 0 auto;
+    }
+    .introduction h1{
+     margin-top: 20rem ;
+     letter-spacing: 5px;
+     font-weight: 2000px;
```
ChrisOwen101: This changes how *bold* the text is and has a maximum value of 900. I don't think you need this here :)

css/style.css
```diff
@@ -1,19 +1,140 @@
...
+        border-radius: 5px;  
+    }
+    .cases {
+        background-color: whitesmoke;
+        padding-bottom: 7px;
+        padding-top: 5px;
+    }
+    .cases_h1{
+        text-align: center;
+        font-size: x-large;
```
ChrisOwen101: It's normally better to define `font-size` in `px`

e.g. `font-size:120px;`

css/style.css
```diff
@@ -1,19 +1,140 @@
...
+        list-style-type: none;
+        display: flex;
+        
+    }
+    .social_media li img {
+        width: 50px;
+    }
+
+
+    @media (max-width: 328px){
```
ChrisOwen101: It'd be great if you could carry on working on the media queries so it scales nicely!

PR #48, opened 2020-02-04
Alex b

AlexandruBudaca:


index.html
```diff
@@ -3,19 +3,87 @@
...
+            <a href="#" class="fa fa-instagram"></a>
+            
+        </div>
+       <p>&copy; Alexandru Budaca</p> 
+    </footer>
+   
+   
+
+
+    </section>
```
LucyMac: This `<section>` wraps part of the main body of the page, and also the footer. So your whole 'features' section is neither in `main` nor in `footer`.
Try moving the section to be inside the `<main>` and keep the footer separate?

css/style.css
```diff
@@ -1,19 +1,223 @@
...
+
+/*MAIN*/
+img {
+    display: block;
+}
+
+#back {
+    display: flex;
+    justify-content: center;
+    background-image: url('/img/snow_background copy.jpg');
```
LucyMac: Try and fix your image by editing the path here. 
Hint: try adding a dot or 2 at the beginning 😉

css/style.css
```diff
@@ -1,19 +1,223 @@
...
+    height: 100vh;
+    background-repeat: no-repeat;
+    background-size: cover;
+    background-position: center;
+}
+
+#title {
```
LucyMac: So many IDs! 😭 
Use classes only! (see my comment below)

css/style.css
```diff
@@ -1,19 +1,223 @@
...
+#social {
+    display: flex;
+    justify-content: center;
+    
+}
+
+.fa {
+    padding: 10px;
+    font-size: 20px;
+    width: 20px;
```
LucyMac: Try adding a height here to make these round

css/style.css
```diff
@@ -1,19 +1,223 @@
...
+    font-size: 40px;
+}
+
+#features {
+    display: flex;
+    justify-content: center;
+    flex-wrap: wrap;
+    align-items: center;
+}
+
```
LucyMac: Can you add a media query to make the 'features' stack vertically on mobile _without wrapping first_. At the moment they just wrap to fit the page size, but it would be good to have a little more control and decide when they should be vertical and when they should be horizontal.

PR #49, opened 2020-02-05
flexbox/karma/TaniaNM

tanianm2017:


css/style.css
```diff
@@ -1,9 +1,213 @@
...
+
+#active-link {
+    font-weight: bold;
+    color: black;
+}
+
+/*Hero styles*/
+
+.hero-image {
+    width: 1980px;
```
LucyMac: Don't set such a precise width here - if you want an element to always be the width of the screen, set it to `width: 100%`

css/style.css
```diff
@@ -1,9 +1,213 @@
...
+    background-color: white;
+    border-bottom: 1px solid grey;
+    top: 0;
+    width: 100%;
+}
+
+#logo {
+    width: 30px;
+    padding-left: 200px;
+    float: left;
```
LucyMac: Can you replace `float: left` with flexbox?
It is not recommended to use floats, they were used before flexbox was created, and are now deprecated.

css/style.css
```diff
@@ -1,9 +1,213 @@
...
+}
+
+#nav-bar {
```
LucyMac: Please never use IDs in CSS! They are not re-usable like CSS classes, and can cause you trouble with clashing styles.
Always prefer to style a class (`.navigation`) or an HTML element (`p`).
I recommend you swap all your IDs for classes, in your CSS but also in your HTML of course.

css/style.css
```diff
@@ -1,9 +1,213 @@
...
+/*Hero styles*/
+
+.hero-image {
+    width: 1980px;
+    height: 735px;
+    background-image: url(../img/first-background.jpg);
+    background-size: cover;
+    background-repeat: no-repeat;
+    background-position: center;
+    position: relative;
```
LucyMac: Good

css/style.css
```diff
@@ -1,9 +1,213 @@
...
+}
+
+
+.hero-text {
+    color: white;
+    text-align: center;
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
```
LucyMac: Ok nice, but can you use Flexbox to do this instead?

PR #50, opened 2020-02-05
Meisam

Erwin6997:


index.html
```diff
@@ -12,10 +12,97 @@
     <link rel="shortcut icon" type="image/x-icon"  href="favicon.ico">
 </head>
 <body>
+<!--navigation bar -->
```
ChrisOwen101: You can format the code in this file to make to easier to read by using the instructions here
https://stackoverflow.com/a/29973358

css/style.css
```diff
@@ -1,19 +1,262 @@
...
+}
+
+.segments{
+    display: flex;
+    justify-content: space-evenly;
+    align-items: baseline;
+    text-align: center;
+}
+
+.nuzzle{
```
ChrisOwen101: All of these three classes have the same contents `width: 60%;`. It'd be better if you had one generic class

```
.segmentImage {
	width: 60%;
}
```
Then you can re-use the code which means you don't have to write so much!

index.html
```diff
@@ -12,10 +12,97 @@
...
+
+ <!-- SEGMENTS -->
+<section class="seg">
+
+ <h2>
+    Everyone needs a little Karma.
+ </h2>
+ <div class="segments">
+    <div class="devices">
+        <img src="http://thinkful-ed.github.io/karma-clone/img/icon-devices.svg" alt="Devices" class="dev ">
```
ChrisOwen101: Rather than using a full URL for this image it's better to use a relative one 

```
../img/icon-devices.svg
````

**Comment by ChrisOwen101** [comment]
Great work!

File: css/style.css [hunk 4]
```diff
@@ -1,19 +1,262 @@
...
+	position: relative;
+}
+li {
+	list-style: none;
+}
+a {
+	color: Black;
+	text-decoration: none;
+}
+.container {
```
ChrisOwen101: Test

Erwin6997:
thanks i fixed

PR #51, opened 2020-02-05
Flexbox/karma/melanie titley

melanieet:
This is very much a work-in-progress. Having missed the classes, I'm not confident that I'm using flexbox correctly. Any feedback would be gratefully received!

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -6,10 +6,6 @@ body {
     -webkit-font-smoothing: antialiased;
 }
 
-.navbar {
-   display: flex; 
-   list-style: none;
-}
 
 /**
  * Add your custom styles below
@@ -21,4 +17,12 @@ body {
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
 
+.meet-karma {
+    background-image: url(../img/first-background.jpg);
+}
 
+ul {
+    display: flex;
+    list-style: none;
+    justify-content: space-evenly;
+}
--- a/css/style.css
+++ b/css/style.css
@@ -70,6 +70,7 @@ ul {
     font-size: 45px;   
 }
 
+
 .about li {
     display: flex;
     flex-wrap: wrap;
--- a/css/style.css
+++ b/css/style.css
@@ -7,13 +7,6 @@ body {
     -webkit-font-smoothing: antialiased;
 }
 
-ul {
-    display: flex;
-    list-style: none;
-    justify-content: space-evenly;
-
-}
-
 /**
  * Add your custom styles below
  * 
@@ -24,10 +17,58 @@ ul {
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
 
+ /* Header styles */
+ .header {
+     display: flex;
+     position: fixed;
+     background-color: white;
+     padding-bottom: 5px;
+     width: 100%;
+     border-bottom: 2px solid rgb(234, 235, 236);
+     ;
+ }
+
+ .header-img {
+     padding-top: 12px;
+     padding-left: 275px;
+     padding-right: 185px;
+ }
+
+ /* Navigation styles */
+ .links .active a {
+     font-weight: 500;
+     font-size: 17px;
+     text-decoration: none;
+     color: rgb(76, 80, 88);
+     }
+
+ a {
+    text-decoration: none;
+    font-size: 17px;
+    color: rgb(131, 137, 148);
+ }
+
+ .navigation ul {
+     display: flex;
+     flex-direction: row;
+     list-style: none;
+     padding-top: 10px;
+ }
+
+ .navigation li {
+     display: flex;
+     padding-right: 30px;
+     color: darkgrey;
+ }
+
+ .navigation .links a:hover {
+     color: rgb(32, 27, 27);
+ }
  
+ /* Introductory section styles */
 .karma-introduction {
     background-image: url(../img/first-background.jpg);
-    background-position: center bottom ;
+    background-position: center bottom;
     padding: 250px;
     text-align: center;
 }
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -16,4 +17,182 @@ body {
...
+ }
 
+ .header-img {
+     padding-top: 12px;
+     padding-left: 275px;
+     padding-right: 185px;
+ }
+
+ /* Navigation styles */
+ .links .active a {
```
jamesgorrie: Often in CSS it can be best to give elements a class name rather than referencing the element type.

e.g.
`.links .active a` will match
```HTML
<div class="links">
  <div class="active">
    <a>The link</a> <!-- <= this matches -->
  </div>
</div>
```

And also
```HTML
<div class="links">
  <div class="active">
    <a>The link</a>  <!-- <= this matches -->
    <div class="sublinks">
      <a>Sublink</a> <!-- <= this will also match -->
    </div>
  </div>
</div>
```

A way to avoid this this would be defining an class to give the element. e.g.
`.links .active .link`

```HTML
<div class="links">
  <div class="active">
    <a class="link">The link</a> <!-- <= this matches -->
  </div>
</div>

<div class="links">
  <div class="active">
    <a class="link">The link</a>  <!-- <= this matches -->
    <div class="sublinks">
      <a class="sublink">Sublink</a> <!-- <= this will not match -->
    </div>
  </div>
</div>
```

jamesgorrie: [Here is an interesting article](https://css-tricks.com/strategies-keeping-css-specificity-low/) about why this might cause issues later in developing the website.

css/style.css
```diff
@@ -16,4 +17,182 @@ body {
...
+     display: flex;
+     position: fixed;
+     background-color: white;
+     padding-bottom: 5px;
+     width: 100%;
+     border-bottom: 2px solid rgb(234, 235, 236);
+     ;
+ }
 
+ .header-img {
```
jamesgorrie: These values seem to move the logo very far to the right - is this what we're looking for?

Screenshot 2020-02-07 at 09 11 52

css/style.css
```diff
@@ -16,4 +17,182 @@ body {
...
+    padding-bottom: 25px;
+}
+
+.about img {
+    padding-bottom: 25px;
+}
+
+.about ul {
+    display: flex;
+    flex-wrap: wrap;
```
jamesgorrie: Do you know why we are allowing this to wrap?

css/style.css
```diff
@@ -16,4 +17,182 @@ body {
...
+    flex-direction: row;
```
jamesgorrie: `flex-direction` is defaulted to row, so you don't need to include it - but you can if it makes things more clear.

css/style.css
```diff
@@ -16,4 +17,182 @@ body {
...
+    padding-top: 20px;
+    font-size: 15px;
+}
+
+.footer ul {
+    padding-left: 10px;
+}
+
+.footer li{
+    display: inline;
```
jamesgorrie: We could use flex here too instead? Do you have an idea of why you might use one over the other?

css/style.css
```diff
@@ -16,4 +17,182 @@ body {
...
+    padding-left: 15px;
+}
+
+.instagram {
+    color: rgb(63, 114, 155);
+    padding: 10px;
+    padding-bottom: 10px;
+    padding-right: 12px;
+    padding-left: 12px;
+}
```
jamesgorrie: `.twitter`, `.facebook`, and `.instagram` => As these are similar could we create 1 class for them ?

index.html
```diff
@@ -7,15 +7,61 @@
...
+            <h2>Bring WiFi with you, everywhere you go.</h2>
+
+            <button type="button">Learn More</button>
+        </section>
+
+        <section class="about">
+            <h1>Everyone needs a little Karma.</h1>
+
+            <ul class="benefits">
+                <li><img width="127" height="127" src="img/icon-devices.svg" alt="Laptop, tablet and phone icon">
```
jamesgorrie: Alt text here might not be useful as we have the text with the description anyway. Generally it is really good to have though!

[There's a good article here](https://blog.hubspot.com/marketing/image-alt-text) giving good and bad examples.

index.html
```diff
@@ -7,15 +7,61 @@
...
+        <hr>
 
-    <!-- Add your HTML markup here -->
-    <!-- Remember: Use semantic HTML tags like <header>, <main>, <nav>, <footer>, <section> etc -->
-    <!-- All the images you need are in the 'img' folder -->
+        <footer class="footer">
+            <h5>Join us on</h5>
+            
+            <ul>
+                <li class="twitter"><i class="fa fa-twitter"></i></li>
```
jamesgorrie: What is the `<i>` tag for?

Did we write any CSS for the `.fa` class?

css/style.css
```diff
@@ -16,4 +17,182 @@ body {
...
+
+.footer li{
+    display: inline;
+    list-style: none;
+    border: 1px solid rgb(234, 235, 236);
+    border-radius: 50%;
+    margin: 5px;
+    cursor: pointer;
+}
```
LucyMac: Try giving a fixed width and height to these 3 elements to ensure the circle around them is fully round.
Another thing that's interfering with the 'roundness' is that the padding you've set on the 3 classes below is not equal on all sides, so it's stretching the circle.

LucyMac:
Hi Melanie, nice work! Can you try and add 2-3 media queries to make it scale better on mobile?

css/style.css
melanieet: I opened the Karma clone website in one window and my copy in another window directly underneath. I attempted to line up the logo as best I could. I'm not sure if this looks different in your browser, @jamesgorrie?

css/style.css
melanieet: Removed, thank you.

css/style.css
melanieet: Finally managed to achieve rounded borders, those things were tricky. Thank you, @LucyMac.

css/style.css
melanieet: Changed.

PR #55, opened 2020-02-15
KarmaClone

karenrogersn:


css/style.css
```diff
@@ -16,4 +35,83 @@ body {
...
+body {
+    line-height: 1;
+}
+
+
+/* Nav section styling */
+
+.navigation {
+    width: 100%;
+    background-color: pink;
```
LucyMac: Remember to remove the pink colour 😉

css/style.css
```diff
@@ -16,4 +35,83 @@ body {
...
+    height: 45.938rem;
+}
+
+
+/* Boxes styling */
+
+.features {
+    display: flex;
+    justify-content: center;
+}
```
LucyMac: This is good! Can you add a media query for these list-items inside to appear in a column on mobile?

A media query looks like this: 
``` 
@media (max-width: 700px) {
    .features {
      // rule to make it into a column here (use flexbox!)
    }
  }
```

css/style.css
```diff
@@ -16,4 +35,83 @@ body {
...
+
+.caption {
+    position: relative;
+    top: 150px;
+    right: 160px;
+    width: 200px;
+    text-align: center;
```
LucyMac: This rule is causing the `caption`s to appear outside of their parent. Remove this rule, move the captions inside their respective list items, and position them using flexbox.

css/style.css
LucyMac: In this particular example, this media query would set a CSS rule that would apply _only_ when the screen is smaller than 700px wide. 
In other words, the CSS you've already written would work on desktop (leave it as it is), and the CSS inside this media query would work on screens smaller than 700px.

index.html
```diff
@@ -20,58 +19,61 @@
...
+                    <li><img width="127" height="127" src="img/icon-devices.svg" alt="electronic devices image" />
+                        <p class="caption">Internet for all devices</p>
+                    </li>
+                    <li><img width="127" height="127" src="img/icon-coffee.svg" alt="cup of coffee image" />
+                        <p class="caption">Boost your productivity</p>
+                    </li>
+                    <li><img width="127" height="127" src="img/icon-refill.svg" alt="gas refilling device image" />
+                        <p class="caption">Pay as You Go</p>
+                    </li>
+                </ul>
```
LucyMac: Perfect structure!

index.html
```diff
@@ -20,58 +19,61 @@
...
-                    <h3>Pay as You Go</h3>
-                </div>
-            </ul>
+            <div class="content">
+                <h2>Everyone needs a little Karma.</h2>
+                <ul class="features">
+                    <li><img width="127" height="127" src="img/icon-devices.svg" alt="electronic devices image" />
+                        <p class="caption">Internet for all devices</p>
+                    </li>
+                    <li><img width="127" height="127" src="img/icon-coffee.svg" alt="cup of coffee image" />
```
LucyMac: In general, avoid using `width` and `height` attributes on your `img` tags. Handle image sizes directly in your CSS file.

css/style.css
```diff
@@ -15,103 +13,192 @@ body {
...
+    }
+    /* Responsive Design */
+    
+    @media screen and (max-width: 700px) {
+        .features {
+            display: flex;
+            flex-direction: column;
+            align-items: center;
+        }
+    }
```
LucyMac: Nice! 🙂

css/style.css
```diff
@@ -15,103 +13,192 @@ body {
...
+    h3 {
+        font-size: 1.7em;
+        margin-top: 0.2em;
+        margin-bottom: 1.2em;
+    }
+    /* Boxes styling */
+    
+    .boxes {
+        position: relative;
+        left: 2em;
```
LucyMac: This rule is making that section mis-align on mobile. Remove it, and add flexbox rules to each `li` to center the icon and text inside each list item.

css/style.css
```diff
@@ -15,103 +13,192 @@ body {
...
+    .features {
+        display: flex;
+        justify-content: center;
+    }
+    
+    .caption {
+        font-size: 1.5em;
+        margin-top: 1em;
+        position: relative;
+        right: 1.8em;
```
LucyMac: Don't rely on this sort of rule when what you want to achieve is centered elements. Remove this rule, and add flexbox rules to the `li` to center its contents (i.e. the image and the caption).

PR #58, opened 2020-02-27
London-Class6/Karma/babak

babakjahangiri:


css/style.css
```diff
@@ -15,5 +15,249 @@ body {
...
+  align-content: space-between;
+  text-align: center;
+  flex-wrap: nowrap;
+  
+}
+
+.content {
+  padding: 0;
+  margin: 0 auto;
+  width: 25rem;
```
LucyMac: this breaks the layout on mobile

index.html
```diff
@@ -1,21 +1,80 @@
...
+      <div class="content">
+        <div class="cborder">
+          <h1>ICYNICY</h1>
+          <h2>A delicious taste in the world</h2>
+          <div class="buttons">
+            <button>Lick more</button>
+          </div>
+        </div>
+      </div>
+    </section>
```
LucyMac: This probably should go inside `main` instead of the header.

PR #82, opened 2020-11-21
flexbox karma kawthar

kwthr8:
# Your Details

- Your Name: kawsar
- Your City:
- Your Slack Name: kawthar

# Homework Details

- Module:html/css
- Week:2


index.html
```diff
@@ -5,17 +5,77 @@
...
+            <span class="navitems">
+                <a>Help</a>
+            </span>
+            <span class="navitems">
+                <a>Login</a>
+            </span>
+        </nav>
+    </div>
+    <div id="banner">
+        <div style="font-family: 'open_sansregular'; font-size: 3.125rem; font-weight: 300;">Introducing Karma</div>
```
psherl, line 42: Most of this should go in your css file.

psherl (commented):
Visually this is very similar to the screenshot, so well done. This shows an understanding of flexbox and attention to detail. There's only a couple of things like the footer icons and the hover state that look to be different.

I've added several comments about the rest of the CSS. Generally we want to use classes instead of IDs. There's several ways you can make this easier for yourself by being a little more organised and only including files that are used in the project within your pull request. There's also some instances where your css is using properties that don't exist.

A lot of the pointers can and will improve with experience, but the flexbox is generally good in this submission.

PR #94, opened 2020-12-06
London Class-7/Murat Demirtas/HTML-CSS-Coursework-Week2 completed

muratdemirtas2001:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:


css/style.css
```diff
@@ -6,14 +6,207 @@ body {
...
+        
+    }
+    #start-link{
+        margin-right: 20px;
+    }
+
+}
 
+/* adding media query for big screens-which hide the mobile nav bar and links */
+@media screen and (min-width:1500px){
```
LucyMac, line 202: Nice - I suggest reducing your breakpoint as 1500px is already considered a large desktop. Try breaking somewhere between 600px and 800px instead (a common one if 768px because that is the width of an iPad screen)

index.html
```diff
@@ -12,10 +12,115 @@
...
+                    <img src="./img/icon-coffee.svg" alt="picture showing a computer">
+                    <p>Boost your productivity</p>
+                </section>   
+                <section class="container-icons-wrapper">
+                    <img src="./img/icon-refill.svg" alt="picture showing a computer">
+                    <p>Pay as you go</p>
+                </section>       
+            </section>
+        </section>
+        <hr>
```
LucyMac, line 89: Can you add some styling to the `hr` so that it doesn't span across the whole width of the screen?

css/style.css
```diff
@@ -6,14 +6,207 @@ body {
...
+        background-image: url('../img/first-background.jpg');
+        height: 500px;
+        background-repeat: no-repeat;
+        background-position: center;
+        background-size: cover;
+        color:white;
+        display:flex;
+        flex-direction: column;
+        justify-content: center;    
+    }
```
LucyMac, line 142: You've already applied these styles to `.hero` up on line 41, outside of any media query. This means these styles will still apply for this breakpoint, and you don't need to write them twice. 

When using media queries, only include styles that are _different_ from the default styles you have already declared.

css/style.css
```diff
@@ -6,14 +6,207 @@ body {
...
+}
+/* adding media query-inorder not to show navbar for big screens,make visible the mobile nav bar and changing background image height */
+@media screen and (max-width:1500px){
+    nav .big-screen{
+        display: none; 
+    }
+
+    .mobile-screen {
+        display:block;
+      }
```
LucyMac, line 126: I notice if you open this mobile nav, the main image gets pushed down below it. To avoid this, you need to take the mobile nav 'out of the flow of the page' using `position: absolute;` so it sits on top of the image and doesn't disturb the natural flow of the page.

LucyMac (commented):
Nice work, well organised and I like you mobile nav! 🙂 

A few comments to further improve this...

css/style.css
muratdemirtas2001: Thank you for this

css/style.css
muratdemirtas2001: Thank you very much,  I have changed it to 768px

@media screen and (min-width:768px){

index.html
muratdemirtas2001: Thank you very much, I changed its' width to 50%, looks much better now.
#main-hr{
    width:50%;
}

css/style.css
muratdemirtas2001: Thank you very much for advising to use position:absolute and explaining now I understand it very clearly. I changed it as you advise and work perfectly. 
.topnav 
{
        overflow: hidden;
        background-color: rgb(94, 91, 91);
        position: absolute;
        display:none;
        min-width: 100vw;
        right:0px;
        top:65px;
       
      }

PR #97, opened 2020-12-06
London Class 7 - Adem Alagoz - HTML/CSS - Week 2

ademalagoz:
# Your Details

- Your Name: Adem Alagoz
- Your City: London
- Your Slack Name: ademalagoz

# Homework Details

- Module: HTML/CSS 
- Week: 2


css/style.css
```diff
@@ -4,8 +4,251 @@
...
+  {
+  color: rgb(235, 106, 31);
+  }
+
+  li {
+    display: inline;
+    padding: 7px;
+    margin: 10px;
+    margin-top: 0;
+    float: right;
```
LucyMac, line 44: Please don't use `float` rules in your CSS. Float is the old way of making layouts in CSS and it has now been replaced by Flexbox (which was introduced a few years ago). 
Set `display: flex` on the parent of the element (`ul` in this case)

css/style.css
```diff
@@ -4,8 +4,251 @@
...
+    background-size:cover;
+    background-position: center;
+    background-repeat: no-repeat;
+    margin: 0;
+    padding: 0;
+    width: 1500px;
+    height: 1000px;
+    align-items: center; 
+    position: relative;
+    top: 0px;left: 30px;
```
LucyMac, line 59: You shouldn't need these `top` and `left` rules. These are only used to position elements that are not in the normal flow of the page, usually with `position: absolute`. 
Here you should be able to do everything with flexbox. 
Remove all the instances of `top` and `left` in your CSS, and replace fixed widths with percentage widths. For example here, change the width to `width: 100%`

css/style.css
```diff
@@ -4,8 +4,251 @@
...
+    
+    
+  }
+
+.photo h3,
+h4 {
+  color: white;
+  text-align: center;
+  position: relative;
+  top: 270px;left: 200px;
```
LucyMac, line 69: Remove `top` and `left` rules and add `display: flex` on the parent element: `.photo`

css/style.css
```diff
@@ -4,8 +4,251 @@
...
+  top: 360px;left: 15px;
+}
+
+.photo h4 {
+  font-size: 50px;
+  font-weight: 50;
+  margin-top: 50px:
+  position: relative;
+  top: 310px;left: -30px;
+}
```
LucyMac, line 88: Quite a bit of repetition here. 
Again, please remove `top` and `left` rules.

css/style.css
```diff
@@ -4,8 +4,251 @@
...
+  
+  line-height: 2.8;
+  padding: 0 20px;
+  font-size: 1rem;
+  text-align: center;
+  color: #fff;
+  border-radius: 5px;
+  background-color: rgb(209, 115, 52);
+  position: relative;
+  top: -430px;left: 700px;
```
LucyMac, line 100: Remove

css/style.css
```diff
@@ -4,8 +4,251 @@
...
+  
+  }
+footer img{
+  
+    width: 15px;
+    height: 15px;
+    margin: 10px;
+    margin-top: 5px;
+    position: relative;
+    left:725px;
```
LucyMac, line 136: Remove

css/style.css
```diff
@@ -4,8 +4,251 @@
...
+    height: 30px;
+    position: relative;
+    top: 10px;left: 15px;
+}
+  .navbar {
+    display: none;
+  }
+.hamburger{
+  display: block;
+  float: right;
```
LucyMac, line 159: Don't use float

css/style.css
```diff
@@ -4,8 +4,251 @@
...
+
+}
+.photo{
+  background-image: url(/img/first-background.jpg);
+  background-size:cover;
+  background-position: center;
+  background-repeat: no-repeat;
+  width: 100%;
+  height: 600px;
+  position: relative;
```
LucyMac, line 170: You don't need to repeat rules that are the same as for desktop. Only put what is _different_ in your media query (for example the height)

index.html
```diff
@@ -12,10 +12,76 @@
...
+        <ul>  
+            <li> <a href="#login"class="navbar">Login</a></li>
+            <li> <a href="#help" class="navbar">Help</a></li>
+            <li> <a href="#help" class="navbar">Blog</a></li>
+            <li> <a href="#store"class="navbar">Store</a></li>
+            <li> <a href="#how it works"class="navbar">How It Works</a></li>
+            <li> <b><a href="#meet karma"class="navbar" >Meet Karma</a> </b></li>
+      
+        </ul>
+
```
LucyMac, line 37: This should live inside the `header`

index.html
```diff
@@ -12,10 +12,76 @@
...
+           
+
+    </section>
+    <button class="learn"
+        type="button">
+    Learn More
+</button>
+    
+    <section class="icon">
+       <p>Everyone needs a little Karma.</p>
```
LucyMac, line 52: This could probably be a heading tag (`h1-6`)

index.html
```diff
@@ -12,10 +12,76 @@
...
+        <img src="img/icon-devices.svg" alt="device">
+        <img src="img/icon-coffee.svg" alt="coffee">
+        <img src="img/icon-refill.svg" alt="refill">
+        
+    </section>
+    <section class="specs">
+        <p>Internet for all devices</p>
+        <p>Boost your productivity</p>
+        <p>Pay as You Go</p>
+
```
LucyMac, line 62: Move these descriptions so that they are with their respective image. For example:
```
       <div class="icon-wrapper">
           <img src="img/icon-devices.svg" alt="device">
           <p>Internet for all devices</p>
       </div>
``` 
(for each icon)

LucyMac (commented):
Good start, your HTML structure is good but your CSS is more complicated than it needs to be. 
Please remove all instances of `top` and `left` in your CSS, and replace with flexbox (remember `display: flex` must be on the parent element). 
To help you fix your layout you can add temporary borders to your elements so you can see them on the page. 
Make a few changes and then I'll have a look again 👍

ademalagoz:
All done Lucy

PR #98, opened 2020-12-06
added main tags

aishak05:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

### DIFF
--- a/index.html
+++ b/index.html
@@ -14,7 +14,7 @@
 <body>
     <header>
         <nav>
-            <img src="http://www.w3.org/2000/svg" alt="karma logo" class="karma logo">
+            <img src="img/karma-logo.svg" alt="karma logo" class="karma-logo">
                 <ul class="navbar">
                     <li><a href="#">Meet Karma </a></li>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -12,6 +12,27 @@
...
+                    <li><a href="#">Store</a></li>
+                    <li><a href="#">Blog</a></li>
+                    <li><a href="#">Help</a></li>
+                    <li><a href="#">Login</a></li>
+                </ul>  
+        </nav>
+    </header>
+
+    <main>
+        <img src="img/first-background.jpg" alt="background image" class="background image">
```
LucyMac, line 30: You'll be better off using a background image here instead of an actual `img` tag. This means, instead of using `img` in your HTML, you can have a more generic tag like `section` for example, and then in your CSS for that section, you can use the rule `background-image` and assign it the image. 
We use this technique when the image is not an essential part of the content, and when we want to overlay text on top of the image.
Details here: https://www.w3schools.com/cssref/pr_background-image.asp

PR #107, opened 2020-12-09
class7-Ali nosratipour 

alinosratipour:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:


css/style.css
```diff
@@ -1,19 +1,171 @@
...
+/* Three Icons Section*/
+
+.icon{
+width:150px ;
+}
+
+.icons{
+font-size: 1.4rem;
+display: flex; 
+gap:50px;
```
LucyMac, line 132: Watch out, `gap` does not work in all browsers.

css/style.css
```diff
@@ -1,19 +1,171 @@
-
- /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
+@import url('https://fonts.googleapis.com/css2?family=Open+Sans+Condensed:wght@300&family=Rajdhani:wght@300&display=swap');
+/* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
```
LucyMac, line 2: Fine to experiment in this context, but remember when copying an actual design, you will be expected to use the required font - in this case, Roboto.

css/style.css
```diff
@@ -1,19 +1,171 @@
...
+padding-bottom: 90px;
+}
+
+/*container that hold Text and button link over the image*/
+.image_text_button{
+position: absolute;
+top: 30%;
+left: 30%;
+transform: translate(-30%, -30%);
+}
```
LucyMac, line 77: It's best to avoid using `position: absolute` unless it's really necessary - because it can cause complications with layouts. 
In this case, you don't need to use it. If you set the parent element (`.image_section`) to `display: flex` and add flex rules to make the content centered, plus a bit of padding inside - you'll see your section should grow and have the text nice and centered.

LucyMac: (this will work once you remove your `img` - see comment in your HTML file)

LucyMac (commented):
Good start Ali - I've added a few comments to help you solve your main image problem. 
Give it a go and let me know once you're ready for another review 🙂

alinosratipour (commented):
Hello, Lucy thank you for your valuable feedbacks  I followed your direction and made some change to webpage. can you kindly give me a feedback.
kind regards

PR #110, opened 2020-12-09
Birmingham Class 2 - Haroon - HTML/CSS - Week 2

psherl:
# Your Details

- Your Name: Haroon Iqbal
- Your City: Birmigham
- Your Slack Name:

# Homework Details

- Module:
- Week:

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -2,3 +2,66 @@ body {
   font-family: "Roboto", sans-serif;
   -webkit-font-smoothing: antialiased;
 }
+
+/* header (nav) */
+
+.nav_bar {
+  background-color: white;
+  position: fixed;
+  width: 100%;
+}
+
+.nav_list {
+  float: right;
+  font-size: 1rem;
+  padding-right: 18rem;
+}
+
+.nav_list li {
+  display: inline-block;
+  margin-left: 1rem;
+  color: grey;
+}
+
+.logo_nav {
+  padding-left: 20rem;
+}
+
+.nav_list .highlighted {
+  font-weight: 500;
+  color: rgb(75, 73, 73);
+}
+/* intro main */
+
+.intro {
+  background-image: url("../img/first-background.jpg");
+  background-position: center;
+  height: 45rem;
+}
+
+.intro h1 {
+  color: white;
+  text-align: center;
+  padding-top: 10rem;
+  font-size: 3rem;
+  font-weight: 300;
+}
+
+.intro h3 {
+  color: white;
+  text-align: center;
+  font-size: 1.5rem;
+  font-weight: 300;
+}
+
+/* Button */
+
+.button {
+  text-align: center;
+  display: inline-block;
+  font-size: 1.125rem;
+  background-color: orange;
+  border-color: orange;
+  border-radius: 4px;
+  padding: 1rem 1.5rem;
+}
--- a/css/style.css
+++ b/css/style.css
@@ -9,10 +9,12 @@ body {
   background-color: white;
   position: fixed;
   width: 100%;
+  height: 4.5rem;
   display: flex;
   flex-direction: row;
   justify-content: space-around;
   align-items: center;
+  top: 0;
 }
 
 .nav_logo {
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -1,19 +1,100 @@
...
-    font-family: 'Roboto', sans-serif;
-    -webkit-font-smoothing: antialiased;
+.nav_bar {
+  background-color: white;
+  position: fixed;
+  width: 100%;
+  height: 4.5rem;
+  display: flex;
+  flex-direction: row;
+  justify-content: space-around;
```
psherl, line 15: Probably space between would have been more appropriate, and then using padding to move the items away from the edge.

css/style.css
```diff
@@ -1,19 +1,100 @@
...
+.intro h3 {
+  color: white;
+  text-align: center;
+  font-size: 1.5rem;
+  font-weight: 300;
+}
 
+/* Button */
 
+.button {
```
psherl, line 65: In this case, you probably wanted to apply the flex properties to the element that was containing the button.

css/style.css
```diff
@@ -1,19 +1,100 @@
...
+}
+
+/* Features */
+
+.devices {
+  list-style-type: none;
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+}
```
psherl, line 84: For each of the child <li> of the devices, you'd also want to add a flex property with the direction of column to keep everything centred.

css/style.css
```diff
@@ -1,19 +1,100 @@
...
+  padding: 1rem 1.5rem;
+}
+
+/* Features */
+
+.devices {
+  list-style-type: none;
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
```
psherl, line 83: In this case, maybe space around could have been more appropriate.

index.html
```diff
@@ -6,16 +6,96 @@
     <title>Karma</title>
     <link href='//fonts.googleapis.com/css?family=Roboto:400,500,300' rel='stylesheet' type='text/css'>
     <link href="//maxcdn.bootstrapcdn.com/font-awesome/4.2.0/css/font-awesome.min.css" rel="stylesheet">
-    <link rel="stylesheet" href="css/normalize.css">
-    <!-- Add a link to your CSS file here (use the line above to guide you) -->
+    <link rel="stylesheet" href="css/style.css">
```
psherl, line 9: You've got two of the same style.css files being added.

index.html
```diff
@@ -6,16 +6,96 @@
...
+                <div class="button">
+                    <button>Learn More</button>
+                </div>
+            </div>
+        </section>
+
+        <section class="features">
+            <div class="content">
+                
+                <div class = "feature_title">
```
psherl, line 47: watch out for inconsistencies in spacing for your class names

css/style.css
```diff
@@ -1,19 +1,100 @@
...
+.nav_list .highlighted {
+  font-weight: 500;
+  color: rgb(75, 73, 73);
+}
+/* intro main */
+
+.intro {
+  background-image: url("../img/first-background.jpg");
+  background-position: center;
+  height: 45rem;
```
psherl, line 45: Depending on which device you're using, an absolute value of 45rem might not always be appropriate. You could use a percentage or viewport units instead.

css/style.css
```diff
@@ -1,19 +1,100 @@
...
+
+.feature_title {
+  display: flex;
+  justify-content: center;
+}
+
+.caption {
+  font-size: 1.1rem;
+  justify-content: center;
+  font-weight: 2;
```
psherl, line 99: Is font weight 2 a property? Is 500, 700 or bold or regular an option? https://www.w3schools.com/cssref/pr_font_weight.asp

psherl (commented):
There's some evidence of proficiency in flexbox here, generally the HTML is well organised and the commit messages are informative.

There's a bit of confusion with the button where the flex should have been applied to the parent, and the feature items where flex columns would have made the layout a bit neater.

Other than that, being concise and tidy are good skills so well done for keeping this pull request organised and clean.

PR #112, opened 2020-12-09
London Class 7-Yasemin Bayraktutan-HTML/CSS-Week 1

ybayraktutan:
first draft

# Your Details

- Your Name: Yasemin Bayraktutan
- Your City: London Class 7
- Your Slack Name: Yasemin Bayraktutan

# Homework Details

- Module: HTML-CSS
- Week: 1


css/style.css
```diff
@@ -1,11 +1,174 @@
...
     -webkit-font-smoothing: antialiased;
+    position: relative;
+}
+ /* Header  */
+ header {
+    font-size: 24px;
+    width: 100%;
+    height: 20px;
+    padding: 40px;
+    margin: 20px;
```
LucyMac, line 15: This spacing is being counted on top of the `width: 100%` which is why if you scroll right, there is additional space on the side of your page. 
I would remove the margin as you shouldn't need it. 
For the padding, you can add the rule `box-sizing: border-box` to your header which will force the padding to be _included_ in the `width: 100%`. It's quite common to add this rule to every element, by using the star selector:
```
* {
  box-sizing: border-box;
}
```

css/style.css
```diff
@@ -1,11 +1,174 @@
...
+}
+.navbar ul li {
+    list-style-type: none;
+    margin: 10px;
+    padding: 5px;
+
+}
+.navbar ul li a {
+    /* list-style-type: none; */
+    text-decoration: none;
```
LucyMac, line 45: It looks like you need to set a `color` rule for these links. They come out purple for me.

css/style.css
```diff
@@ -1,11 +1,174 @@
...
+.picture-text-heading {
+    font-size: 50px;
+}
+
+.picture-text-smalltext {
+    font-size: 30px;
+}
+
+.picture-button {
+    font-size: 24;
```
LucyMac, line 78: This needs a unit - either 24px or 1.5 rem

LucyMac: Actually, this is a div so remove the font-size completely here and apply it below to the button.

css/style.css
```diff
@@ -1,11 +1,174 @@
...
+    
+}
+
+.picture-button button {
+    width: 150px;
+    height: 50px;
+    padding: 15px 30px;
+    background-color: orangered;
+    text-align: center;
+    font-size: 24;
```
LucyMac, line 88: add unit here

LucyMac: Also, could you add some rounded corners to follow the design a little more closely?
If you add a border of the same orangered colour as the background, it will hide the current default border as well.

css/style.css
```diff
@@ -1,11 +1,174 @@
...
+
+.wrapper-icon-content img {
+    width: 150px;
+    height: 150px;
+}
+
+.wrapper-icon-content p {
+    font-size: 24px;
+    padding: 15px 40px;
+
```
LucyMac, line 134: You could add `text-align: center;` here

css/style.css
```diff
@@ -1,11 +1,174 @@
...
+    text-align: center;
+    margin: 10px;
+    padding: 10px;
+}
+.wrapper-icon {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-around;
+    align-items: center;
+    flex-wrap: wrap;
```
LucyMac, line 114: Can you add media queries to control position these elements on mobile? 
Flex-wrap is a great 'fall-back', and very handy when we have many items on several rows and we want them to flow naturally whatever the screen size. In our case, we want a little more control so it's best to use `flex-direction` in a media query.

LucyMac (commented):
Really nicely organised code Yasemin, and a good use of CSS. 
I've added a few comments for improvements, and also remember to add some media queries to adapt the page for mobile 🙂

index.html
```diff
@@ -12,7 +12,69 @@
     <link rel="shortcut icon" type="image/x-icon"  href="favicon.ico">
 </head>
 <body>
+    <header>
```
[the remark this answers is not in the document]
ybayraktutan: Thank you very much

css/style.css
ybayraktutan: I changed it, thank you.

css/style.css
ybayraktutan: I fixed this problem.

css/style.css
ybayraktutan: I have noticed this problem while adding media quarry and fixed it.

css/style.css
ybayraktutan: I have changed these, thank you.

css/style.css
ybayraktutan: I added this.

css/style.css
ybayraktutan: I added this.

index.html
```diff
@@ -12,7 +12,69 @@
...
+    <hr>
+    <footer>
+        <section class="footer-heading">
+            <p>Join us on</p>
+        </section>
+        <section class="footer-icons">
+            <img src="img/twitter-icon.svg" alt="Twitter Icon"/>
+            <img src="img/facebook-icon.svg" alt="Facebook Icon"/>
+            <img src="img/instagram-icon.svg" alt="Instagram Icon"/>
+        </section>
```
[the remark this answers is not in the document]
ybayraktutan: I fixed this problem. Thank you.

PR #121, opened 2020-12-11
London Class 7 - Somayeh Karimi - HTML/CSS - Week 1

DibaKarimi:
# Your Details

- Your Name:Somayeh Karimi
- Your City:London
- Your Slack Name: Somayeh

# Homework Details

- Module: HTML/CSS
- Week: 1


clairesmi:
Hi Somayeh, this looks really good, great job! 
Here are a few points for you:
HTML - really well structured overall, and good use of semantic elements (particularly your nav element, list ad footers) on line 51, you have a p tag that could be in a section of it's own - I would also consider making this ah2 as it represents what I would consider the second most important heading on the page. Likewise, I would make line 46 a h1 as it is the main heading on the page.

CSS - I noticed your background image isn't showing, this was because when you are linking to a file, you need to use the relative path to that file from the file you're writing in, so that would look like this, meaning that you are directing the browser up a level from your style.css file and then down a level again into your images folder. ```..``` means moving up a level in the file structure and ```/``` means moving down a level. 
```background: url('../img/first-background.jpg') ``` 

From line 152, you used flex-box really nicely to display that section, so try to use flexbox consistently across the site. Be careful with using ```position: absolute``` unless totally necessary as it removes the element from the flow of the document and so doesn't always behave as you want it to. Also - good use of commenting what the different sections of your stylesheet were, that's a good habit to get into :) 

Let me know on slack if you've got any questions about any of this.

**Comment by DibaKarimi** [comment]
Hi Claire,
Thank you for your feedback. I took them on board.
Also, I tried the Live Share and and the background image was showing, however I applied **..**  .

File: css/style.css [hunk 1]
```diff
@@ -1,19 +1,262 @@
...
+      
+      .dropdown:hover .drop-img {
+        background-color: #caceca;
+      }
+      .icon{
+          margin-left: 1px;
+          margin-bottom: 10px;
+      }
+  }
+
```
Buroni, line 107: Nice use of comments to break up your CSS

css/style.css
```diff
@@ -1,19 +1,262 @@
...
+.sec-background{
+    background: url('../img/first-background.jpg')  no-repeat center ;
+    background-size: cover;
+    background-clip: content-box;
+    width: 100%;
+    height: 735px;
+    color: white;
+    text-align: center;
+}
```
Buroni, line 118: `.dv-background` isn't used anywhere -- make sure you remove code that isn't being used

css/style.css
```diff
@@ -1,19 +1,262 @@
...
+.dv-background{
+   text-align: center;
+   transform: translate(-50%, -50%);
+
+}
```
Buroni, line 123: Try not to include the element name in your CSS class -- for example use `.intro` instead of `.p-intro`. 

If you want to restrict `.intro` to only `p` elements, you can instead use `p .intro`:

```css
p .intro {
    font-size: 3.5rem;
    /* ... */
}
```

css/style.css
```diff
@@ -1,19 +1,262 @@
...
     -webkit-font-smoothing: antialiased;
+    width: 100%;
+}
+header{
+    width: 90%;
+    margin-left: 20px;
+}
+
+/******   Navigation styles  *******/
+
```
Buroni, line 18: There's nothing wrong with using `float`, but try to think if you can solve the problem using only `flex` first.

css/style.css
```diff
@@ -1,19 +1,262 @@
...
+    font-weight: 150;
+    margin-bottom: 1px;
+    padding-top: 20%;
+}
+.p-bring{
+    font-size: 1.9rem;
+    font-weight: 100;
+}
+
+input{
```
Buroni, line 135: Do you want to apply this to all `input` elements or just buttons? If just buttons, you should use `input[type=button]` instead of `input`.

css/style.css
```diff
@@ -1,19 +1,262 @@
...
+        color: black;
+        padding: 12px 16px;
+        text-decoration: none;
+        display: block;
+      }
+      
+      .dropdown-content a:hover {
+          background-color: #f1f1f1
+        }
+
```
Buroni, line 94: Good understanding of CSS selectors and `display` in implementing your mobile dropdown!

Buroni (commented):
Great job Somayeh -- the layout looks great, the mobile responsiveness works, and you've shown good understanding of CSS selectors like `hover:` combined with `display` to show/hide the menu.

I've added some suggestions for improvements in your CSS.

There's also one minor layout issue: as you can see in my screenshot, the menu items aren't centre-aligned, but slightly nearer the top of the page. It may be useful to see how you could fix that!

Screenshot 2021-01-17 at 10 54 42

PR #123, opened 2020-12-11
Flexbox/karma/kossivi essilivi

Kos0606:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -37,10 +37,16 @@ h2 {
   margin-top: 20px;
 }
 
-button {
+.amingo {
   display: flex;
   justify-content: center;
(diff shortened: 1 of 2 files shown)


pddys (commented):
Well done, there are some examples here that demonstrate that you can use flexbox.

Overall what would be helpful is to start using classes more, and using IDs less and then you'll be able to keep your code neater and avoid even repeating yourself. You could also save yourself some flexbox issue with better structured HTML. This will come with experience.

Otherwise, it's similar to the image in the exercise so good work!

PR #132, opened 2020-12-12
London class 7/ahmad janah

AhmadJanah:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:


css/style.css
```diff
@@ -1,9 +1,203 @@
...
-    font-family: 'Roboto', sans-serif;
-    -webkit-font-smoothing: antialiased;
+  font-family: "Roboto", sans-serif;
+  -webkit-font-smoothing: antialiased;
+  display: flex;
+  flex-direction: column;
+  padding-left: 5%;
+  padding-right: 5%;
+  margin-left: 5%;
+  margin-right: 5%;
```
LucyMac, line 11: Avoid setting rules like this on the `body` as it will affect your whole page. Start styling from one level down: on the header, main and footer.

css/style.css
```diff
@@ -1,9 +1,203 @@
...
+.maindivimage {
+  position: relative;
+  width: 120%;
+  text-align: center;
+  color: white;
+}
+
+.mainimag {
+  max-width: 100%;
+  height: 80%;
```
LucyMac, line 80: When sizing an `img` tag, never give it a width AND a height as this will cause the image to distort. It will try and stretch the image inside the foxed box that you have assigned it and make it look squashed or elongated. 
It's best if you just give the image `width: 100%` for example, and let the height set naturally.

css/style.css
```diff
@@ -1,9 +1,203 @@
...
+  justify-content: space-around;
+  align-items: center;
+  flex-direction: column;
+  margin-top: 0%;
+  margin-bottom: 10%;
+}
+
+.maindivimage {
+  position: relative;
+  width: 120%;
```
LucyMac, line 73: You have added too much inner space on `body` and now you are trying to overwrite that by setting a width that is larger than 100%. This is a sign that something is not working well somewhere. Avoid setting anything to more than 100%, it will only cause you trouble 😉

css/style.css
LucyMac: In this particular case however, I would not use an `img` tag for this section. 
There is a great way to add a background image in CSS, using the `background-image` property. You could add this directly to you `maindivimage` class.

css/style.css
```diff
@@ -1,9 +1,203 @@
...
+  width: 33.33%;
+}
+
+.bimage {
+  width: 30%;
+  align-items: center;
+  transform: translate(120%, 10%);
+}
+
+.underimagetextr {
```
LucyMac, line 145: When naming your classes, you can use 'kebab-case' which is when everything is lower case but you join each word with a hyphen `-`
In this example it would be `under-image-text`

css/style.css
```diff
@@ -1,9 +1,203 @@
...
+  font-size: 28px;
+  font-weight: 500;
+  text-align: center;
+  transform: translate(0%, 70%);
```
LucyMac, line 149: Don't use the `transform` property when you could achieve the same effect with flexbox or padding. 
In this case, `padding-top` should be enough.

css/style.css
```diff
@@ -1,9 +1,203 @@
...
+}
+
+.solid {
+  border-top: 1px solid #bbb;
```
LucyMac, line 153: This `hr` element has no width, which is why it's not appearing on the screen. 
Add a width and it should show 🙂

css/style.css
```diff
@@ -1,9 +1,203 @@
...
+.divfooter {
+  width: 80%;
+}
+
+.fimage {
+  width: 20%;
+  align-items: flex-end;
+  transform: translate(80%, 10%);
+  border-radius: 50%;
+  border: 0.2px solid #bbb;
```
LucyMac, line 191: Avoid decimal points for pixels. Use 1px as the minimum.

css/style.css
```diff
@@ -1,9 +1,203 @@
...
+}
+
+.divfooter {
+  width: 80%;
+}
+
+.fimage {
+  width: 20%;
+  align-items: flex-end;
+  transform: translate(80%, 10%);
```
LucyMac, line 189: Again, you don't need `transform` here, you just need to set your flexbox rules correctly. The rule `align-items: flex-end;` will have no effect here because `fimage` is not a flex container. 
The best solution would be to make `divfooter` a flex container by adding `display: flex` to it, and also adding the flex rules you need to center your images.

LucyMac (commented):
Nice start Ahmad, but a little overcomplicated at times. Try to avoid `transform` for simple layouts like these, and use flexbox and padding instead. 
Also, remove the main image and replace it with a [background image in your CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/background-image) 👌

AhmadJanah:
Thank you Lucy for all your comments, I update my code taking consider all your notes. I hope I succeeded to solve all the issues.

PR #135, opened 2020-12-12
London - Class 7/Oksana Kostenko - HTML/CSS - Week 1

oksygenn:
# Your Details

- Your Name: Oksana Kostenko
- Your City: London
- Your Slack Name: Oksana Kostenko

# Homework Details

- Module: HTML/CSS
- Week:  Week 1


css/style.css
```diff
@@ -1,19 +1,166 @@
...
+}
+
+.hero p {
+  font-size: 1.2rem;
+  padding: 0 65px 20px 65px;
+  line-height: 2rem;
+}
+
+.hero button {
+  width: 120px;
```
LucyMac, line 67: Watch out setting fixed widths on buttons - it's usually not necessary and can cause trouble if your content grows. Buttons will grow naturally with the content inside them, and for extra breathing space it's more common to use padding.

css/style.css
```diff
@@ -1,19 +1,166 @@
...
+  padding: 0.7rem;
+  background-color: rgb(250, 115, 66);
+  border: none;
+  border-radius: 4px;
+}
+
+/* ABOUT */
+.about {
+  color: rgb(82, 80, 80);
```
LucyMac, line 76: Perfectly fine to use `rgb()` for colours, just be aware hex codes are more common (e.g. #333333)

css/style.css
```diff
@@ -1,19 +1,166 @@
...
+  text-align: center;
+  padding: 40px 60px 10px 60px;
+}
+
+.about h2 {
+  font-weight: lighter;
```
LucyMac, line 82: Usually best to be more precise here. You've got font weights 400, 500 and 300 loaded into the project (see line 11 of index.html) so you can use those values, example: `font-weight: 300`

css/style.css
```diff
@@ -1,19 +1,166 @@
...
+
+  .hero h1 {
+    padding-top: 150px;
+  }
+
+  .flex-container {
+    flex-direction: row;
+    justify-content: space-around;
+  }
+}
```
LucyMac, line 166: Awesome responsive CSS!

index.html
```diff
@@ -1,21 +1,83 @@
...
+    <link rel="shortcut icon" type="image/x-icon" href="favicon.ico" />
+  </head>
 
-    <!-- Add your HTML markup here -->
-    <!-- Remember: Use semantic HTML tags like <header>, <main>, <nav>, <footer>, <section> etc -->
-    <!-- All the images you need are in the 'img' folder -->
+  <body>
+    <header>
+      <nav>
+        <img src="img/karma-logo.svg" alt="company logo" class="logo" />
```
LucyMac, line 27: The logo is technically not part of the nav. It should probably be inside the header, but before the nav.

index.html
```diff
@@ -1,21 +1,83 @@
...
+        <div class="img-container">
+          <img src="./img/twitter-icon.svg" alt="twitter icon" />
+        </div>
+        <div class="img-container">
+          <img src="./img/facebook-icon.svg" alt="facebook icon" />
+        </div>
+        <div class="img-container">
+          <img src="./img/instagram-icon.svg" alt="instagram icon" />
+        </div>
+      </section>
```
LucyMac, line 79: This could potentially be a list

LucyMac (commented):
Great looking page Oksana! Raally clean, organised and thought-through code, and the result looks really nice. 
Extra points for writing media queries! 
Just a few comments below...

css/style.css
oksygenn: Thanks Lucy, i will delete the fixed width

css/style.css
oksygenn: good to know, thanks.

css/style.css
oksygenn: Good point.  Will change that also, thank you

index.html
oksygenn: Wasn't sure which tag i should use here so decided to use div. But list makes much more sense! Will change that too :)

PR #136, opened 2020-12-12
London class 7/hozan ali

Hozan94:
# Your Details

- Your Name: Hozan Ali
- Your City: London class 7
- Your Slack Name: Hozan Ali

# Homework Details

- Module: HTML-CSS-Coursework-Week 2
- Week: week 2

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -20,6 +20,7 @@ header {
   justify-content: space-around;
   height: 100px;
 }
+
 #logo {
   height: 75px;
 }
@@ -57,9 +58,11 @@ a:hover {
   justify-content: center;
   align-items: center;
 }
+
 #firstText {
   font-size: 40px;
 }
+
 #learnMore {
   background-color: rgb(255, 86, 24);
   padding: 35px;
@@ -79,6 +82,7 @@ a:hover {
   padding-top: 100px;
   padding-bottom: 50px;
 }
+
 .contents {
   display: flex;
   justify-content: space-evenly;
@@ -86,6 +90,7 @@ a:hover {
   font-weight: 100;
   color: rgb(46, 46, 46);
 }
+
 #content-1 {
   display: flex;
   flex-direction: column;
@@ -108,6 +113,34 @@ a:hover {
   justify-content: space-between;
   height: 400px;
   text-align: center;
+}
+
+footer {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+
+  font-size: 20px;
+
+  margin-top: 50px;
+  margin-bottom: 30px;
+
+  background-color: rgb(245, 245, 245);
+}
+
+#iconsWrapper {
+  padding-left: 0px;
+  margin: 0px;
+}
+
+#icon {
+  height: 35px;
+  width: 35px;
+}
+
+#copyInfo {
+  color: rgb(173, 173, 173);
+  font-size: 15px;
 }
 /* section {
   width: fit-content;
--- a/css/style.css
+++ b/css/style.css
@@ -18,15 +18,32 @@ body {
 header {
   display: flex;
   justify-content: space-around;
+  height: 100px;
 }
+#logo {
+  height: 75px;
+}
+
 li {
   display: inline-block;
   margin: 5px;
   padding: 5px; /* is it better to use padding or margin??? */
 }
+#meetKarma {
+  font-weight: 500;
+}
+
+a {
+  text-decoration: none;
+  color: rgb(92, 92, 92);
+}
+
+a:hover {
+  color: rgb(255, 86, 24);
+}
+
 .intro {
   height: 750px; /* is there another method to fit the img inside the section without setting height */
-
   background-image: url("/img/first-background.jpg");
   background-repeat: no-repeat;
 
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -16,4 +15,178 @@ body {
...
+}
+
+/* 4.4 adding content height, adjusting the display and aligning the texts */
+#content-3 {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  height: 400px;
+  text-align: center;
+}
```
LucyMac, line 140: Avoid using IDs in CSS at all costs! They are not re-usable and are too hard to overwrite. 

You should use CSS classes, and re-use them. Here, you have 3 different IDs containing exactly the same styles. Create just one CSS class, and re-use it 3 times in your HTML.

css/style.css
```diff
@@ -16,4 +15,178 @@ body {
...
+a:hover {
+  color: rgb(255, 86, 24);
+}
+
+/* 2.0 INTRO SECTION TO KARMA */
+
+/* 2.1 adding background-image, decorating font size for texts, and displaying the contents as a column  */
+.intro {
+  height: 750px; /* is there another method to fit the img inside the section without setting height */
+  background-image: url("/home/coder/Documents/GitHub/HTML-CSS-Coursework-Week2/img/first-background.jpg");
```
LucyMac, line 62: This should simply be `url("../img/first-background.jpg")`

index.html
```diff
@@ -8,14 +8,76 @@
...
+    <main>
+        <article class="intro">
+            <p> Introducing Karma </p>
+            <p id="firstText">Bring WiFi with you, everywhere you go.</p>
+            <button id="learnMore"> Learn More </button>             
+        </article>
+       
+        <article class="part-one">           
+            <p>Everyone needs a little Karma</p>
+        </article>
```
LucyMac, line 45: This probably doesn't need to be a separate article. It's more like the header on the 3 icons below it - so you could change it to a `h2` or `h3` tag, and move it into the next section.

LucyMac: Remember, `<p>` is for both text, not for headers/titles.

css/style.css
```diff
@@ -16,4 +15,178 @@ body {
...
+  flex-direction: column;
+  justify-content: space-between;
+  height: 400px;
+  text-align: center;
+}
+
+/* 4.5 adding a fixed height to the image inside this contents, as the default height was smaller than other images */
+#content-3 img {
+  height: 273.6px;
+}
```
LucyMac, line 145: It's probably not a great idea to set a size for one of the icons and not the others, since they need to all looks the same (this is what's causing the Firefox issue).
Once you have one class for your 3 sections, you can set a rule like 
`.contents-icon img { width: 200px; }` which will then give the same size to all 3 icons.

index.html
```diff
@@ -8,14 +8,76 @@
...
+    </main>
+
+    <footer>
+        <p>Join us on</p>
+        
+        <ul id="iconsWrapper">
+            <li><a href="#"><img id="icon" src="img/twitter-icon.svg" alt="Twitter Icon"></a></li>
+            <li><a href="#"><img id="icon" src="img/facebook-icon.svg" alt="Facebook Icon"></a></li>
+            <li><a href="#"><img id="icon" src="img/instagram-icon.svg" alt="Instagram Icon"></a></li>
+        </ul>
```
LucyMac, line 78: Good semantic HTML here.
Again, like everywhere else, don't use an ID. Change it to a class: `.icon`

LucyMac (commented):
Good looking page @Hozan94 
Nice HTML structure, though remember to use appropriate text tags, such as `h1`, `h2`, `h3` etc for headers, not just `p`.
And very important, please avoid IDs in your CSS and replace with classes which you should re-use as much as possible (for example for the large orange icons section, you can re-use styles for each item).

Hozan94:
> Good looking page @Hozan94
> Nice HTML structure, though remember to use appropriate text tags, such as `h1`, `h2`, `h3` etc for headers, not just `p`.
> And very important, please avoid IDs in your CSS and replace with classes which you should re-use as much as possible (for example for the large orange icons section, you can re-use styles for each item).



> Good looking page @Hozan94
> Nice HTML structure, though remember to use appropriate text tags, such as `h1`, `h2`, `h3` etc for headers, not just `p`.
> And very important, please avoid IDs in your CSS and replace with classes which you should re-use as much as possible (for example for the large orange icons section, you can re-use styles for each item).

Hozan94:
> Good looking page @Hozan94
> Nice HTML structure, though remember to use appropriate text tags, such as `h1`, `h2`, `h3` etc for headers, not just `p`.
> And very important, please avoid IDs in your CSS and replace with classes which you should re-use as much as possible (for example for the large orange icons section, you can re-use styles for each item).

@LucyMac your feedback really appreciated.
I was debatting at first wether to add `h` tag at first or not, as the rule of thumb in elements like `section` is to have at least one inside. But used `p` instaed as I wanted my font to look exactly how it looks in the orignial project, so when I used the `h` tag first the font-weight was really thick.

Regarding the IDs, thanks for notifying me of this issue, I will try to resolve them and any other notes you left on this project.

Lastly, I will try to read more about "relative" and "absolute" path as I did not know the problem I had encountered earlier regarding image loading, was related to this issue. 

Thank you again for the feedback.

PR #139, opened 2020-12-12
London-class-7/Laura-M/HTML & CSS - Week 2

Laura-TM:
Work in progress (trying to add flexboxes)

# Your Details

- Your Name: Laura M
- Your City: London
- Your Slack Name: Laura M

# Homework Details

- Module: HTML & CSS
- Week: 2


css/style.css
```diff
@@ -16,4 +16,326 @@ body {
...
+    text-align: center;
+    font-size: 12px;
+    font-weight: 400; 
+    color: lightgrey;  
+    padding: 15px;
+}
+
+@media only screen and (max-width: 600px) {
+    
+    #header {
```
Behnazz, line 223: you can use hamburger menu to hide the links in header and it this way you don't have that white extra space

Behnazz (commented):
Good job. Welldone. 👍 
Media query needs improvement :)
Soe minor comments here and there

css/style.css
```diff
@@ -16,4 +16,326 @@ body {
...
+    a {
+        font-size: 15px;
+        padding: 12px 10px;
+    }
+    
+    #background-image {
+        background-repeat: no-repeat;
+        margin-top: 5px;
+        display: block;
+        width: 100%;
```
Behnazz, line 247:
```suggestion
        background-repeat: no-repeat;
        margin-top: 5px;
        display: block;
        width: 100%;
```
What happens if you comment these? and changes?

css/style.css
```diff
@@ -16,4 +16,326 @@ body {
...
+        height: auto;
+    }
+    
+    h1 {
+        font-size: 20px;
+        margin-top: 70px;
+    }
+    
+    h3 {
```
Behnazz, line 256: The order of using h tags are important. After h1 we use h2 and then with css change the font size
it is better to use order h1 -> h2 -> h3

css/style.css
```diff
@@ -16,4 +16,326 @@ body {
...
+        font-size: 20px;
+    }
+    
+    .feature-container {
+        height: 100px;
+        width: 200px;
+    }
+    
+    .feature-container img {
+        width: 150px;
```
Behnazz, line 282:
```suggestion
        width: 100%;
```

css/style.css
```diff
@@ -16,4 +16,326 @@ body {
...
+        padding: 10px 0 10px 0;
+        margin: 10px 0 10px 0;
+    }
+    
+    #ext-container {
+        flex-basis: 50%;
+    }
+    
+    #sales-container {
+        flex-basis: 50%;
```
Behnazz, line 306:
```suggestion
 align-items: center;
```

Behnazz (commented):
just some more small changes that can make mobile view better

css/style.css
```diff
@@ -16,4 +16,132 @@ body {
...
+}
+
+#nav-bar {
+    overflow: hidden;
+    padding-top: 10px;
+    width: 100%;
+    justify-content: flex-end;
+}
+
+.nav-link {
```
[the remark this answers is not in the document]
Laura-TM: The links do have a hover effect on [header a:hover], can you please check that is what you mean?

css/style.css
```diff
@@ -16,4 +16,132 @@ body {
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
 
+#header {
```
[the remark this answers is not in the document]
Laura-TM: which other tags can  you target without class or id, the ones that are semantic?

css/style.css
```diff
@@ -16,4 +16,132 @@ body {
...
+    max-height: 50px;
+    display: flex;
+    flex-direction: row;
+    width: 100% 
+    z-index: 9999;
+    top: 0;
+    left: 0;
+}
+
+#header-img {
```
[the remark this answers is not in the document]
Laura-TM: I happened to read an article a year ago suggesting IDs are more powerful, I then made the assumption I should only use those, so I would avoid any potential issues when using both IDs and classes. Lucy explained that I should not use only IDs, so I have now corrected this project to only have classes.

css/style.css
```diff
@@ -16,4 +16,132 @@ body {
...
+    padding: 15px 10px 30px 100px;
+    width: 25px;
+    height: 25px;
+    position: relative;
+    justify-content: flex-start;
+}
+
+#nav-bar {
```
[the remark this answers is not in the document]
Laura-TM: From previous readings, IDs can override classes (which is why I wanted to avoid using classes, but I have understood that is not correct). From today's research, I learnt IDs are easier to select with Javascript and can only be used in one element.

css/style.css
Laura-TM: I have added a burger-menu icon, as I don't know how to completely swap them and not hide them when not in use

PR #140, opened 2020-12-13
London class 7-Aisha Khan - HTML & CSS - Week 2

aishak05:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -11,7 +11,7 @@ body {
 }
 
 .navbar {
-  display: inline;
+  float: left;
 }
 /**
  * Add your custom styles below
--- a/css/style.css
+++ b/css/style.css
@@ -19,7 +19,7 @@ body {
 .navbar li {
   display: block;
   width: 120px;
-  float: left;
+  float: right;
 }
 /**
  * Add your custom styles below
--- a/css/style.css
+++ b/css/style.css
@@ -11,7 +11,8 @@ body {
 }
 
 .navbar {
-  float: left;
+  float: right;
+  display: inline;
 }
 /**
  * Add your custom styles below
--- a/css/style.css
+++ b/css/style.css
@@ -1,31 +1,54 @@
 /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
 
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
+}
+
 body {
   font-family: "Roboto", sans-serif;
   -webkit-font-smoothing: antialiased;
 }
 
-.karma-logo {
-  width: 50px;
-  height: 50px;
+header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 30px 10px;
 }
 
-.navbar {
-  list-style: none;
-  margin: 10px;
-  width: 900px;
+.logo {
+  width: 50px;
+  height: 50px;
+  margin-left: 50px;
 }
 
 .navbar li {
-  display: block;
-  width: 200px;
-  float: right;
-  margin-left: 2px;
+  list-style: none;
+  display: inline-block;
+  padding: 0px 20px;
 }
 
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -1,9 +1,199 @@
...
- /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
+/* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
 
 body {
-    font-family: 'Roboto', sans-serif;
-    -webkit-font-smoothing: antialiased;
+  font-family: "Roboto", sans-serif;
+  -webkit-font-smoothing: antialiased;
+  margin: 160px;
+  font-size: 1em;
```
LucyMac, line 7: It's best not to add specific styles like this to the `body` - it's too generic and you'll want to override those styles most of the time. Start adding your styles one level down: on `header`, `main` and `footer`

LucyMac: Also, to achieve [this design](https://raw.githubusercontent.com/CodeYourFuture/HTML-CSS-Coursework-Week2/master/Karma%20-%20desktop%20design.png), we don't want to create all that white space around the page. We want the main background image to 'bleed' off the page. To achieve this, remove the margin from line 6.

css/style.css
```diff
@@ -1,9 +1,199 @@
...
+}
+
+header {
+  display: flex;
+  justify-content: flex-end;
```
LucyMac, line 12: Have you tried `space-between` here? 🙂

css/style.css
```diff
@@ -1,9 +1,199 @@
...
+  color: white;
+  font-weight: lighter;
+  font-size: 45px;
+}
+
+.button {
+  background-color: rgb(250, 99, 11);
+  border: none;
+  padding: 3px 15px;
+  border-radius: 5%;
```
LucyMac, line 68: Nice. Probably best to set the border-radius in pixels, unless you're going for a full circle like in the footer in which case you can use `50%`

index.html
```diff
@@ -12,6 +12,52 @@
...
+    <main>
+        <section class="main">
+           <h1>Introducing Karma</h1>
+           <h2>Bring WIFI with you, everywhere you go</h2>  
+                <div class="button">
+                    <p>Learn More</p>
+                </div>    
+        </section>
+        <p class ="tagline">Everyone needs a little Karma</p>
+        <container class="mobile-view">
```
LucyMac, line 36: I'm afraid the container tag doesn't exist! We speak of containers as a concept, but there is no specific tag for them. You can use a `section` here, and give it a class of `container` instead. 
You can also probably move the 'tagline' inside of it, and make it a heading tag instead of a paragraph.

index.html
```diff
@@ -12,6 +12,52 @@
...
+            <div class="options">
+                <img src="img/icon-devices.svg" alt="devices" class="devices">
+                <img src="img/icon-coffee.svg" alt="coffee cup" class="coffee-cup">
+                <img src="img/icon-refill.svg" alt="refill icon" class="refill">
+            </div>
+                <div class="text">
+                    <p class="devices-text">Internet for all devices</p>
+                    <p class="coffee-text">Boost your productivity</p>
+                    <p class="refill-text">Pay as you go</p>
+                </div>
```
LucyMac, line 46: These elements should be grouped differently so that each image 'lives' with its text in the same container. Some like this, but 3 times: 
```
                <div class="icon-wrapper">
                    <img src="img/icon-devices.svg" alt="devices" class="devices">
                    <p class="devices-text">Internet for all devices</p>
                </div>
```
This will ensure that in mobile mode, when they are stacked, the images are still next to the corresponding text.

LucyMac (commented):
Nice Aisha, a couple of change to the body margin and changing the structure of the orange icon section should make big improvements.

PR #141, opened 2020-12-13
London Class 7 - Mustafa Ak - HTML/CSS - Week2

akbasmustafa:
# Your Details

- Your Name: Mustafa Ak
- Your City: London
- Your Slack Name: @Mustafa Ak

# Homework Details

- Module: HTML/CSS
- Week: 2


index.html
```diff
@@ -7,15 +7,78 @@
...
+                <li><a href="#">Store</a></li>
+                <li><a href="#">Blog</a></li>
+                <li><a href="#">Help</a></li>
+                <li><a href="#">Login</a></li>
+            </ul>
+        </nav>
+    </header>
+
+    <main>
+        <section id="hero">
```
LucyMac, line 42: Don't use IDs for styling - replace with a class.

index.html
```diff
@@ -7,15 +7,78 @@
...
+            <div>
```
LucyMac, line 43: Since a `div` is a very generic tag, it's best to give it a class. 
Otherwise, if you say `#hero div` in your CSS, it will target all other divs inside `#hero`. I know you don't have any other divs in there at the moment, but it's best practice to be a little more precise to avoid unexpected surprises.

index.html
```diff
@@ -7,15 +7,78 @@
...
+                <p>Bring WIFI with you, everywhere you go.</p>
+                <button>Learn More</button>
+            </div>
+        </section>
+
+        <section id="middle">
+            <p>Everyone needs a little Karma.</p>
+        </section>
+        
+        <section id="boxes">
```
LucyMac, line 54: Again, please change this to a class

index.html
```diff
@@ -7,15 +7,78 @@
...
+            <div>
```
LucyMac, line 55: Do you need this div? It seems to have the same purpose as the `section` (they both wrap the same elements)

index.html
```diff
@@ -7,15 +7,78 @@
...
 
+    <footer>
+        <hr>
+        <p>Join us on</p>
+        <div class="icons">
+            <img class="icon" src="./img/twitter-icon.svg" alt="twitter">
+            <img class="icon" src="./img/facebook-icon.svg" alt="facebook">
+            <img class="icon" src="./img/instagram-icon.svg" alt="instagram">
+        </div>
+        <p style="opacity: 0.7;">&copy; Karma Mobility, Inc.</p>
```
LucyMac, line 81: It's bad practice to use inline styles like this. This is because of [CSS specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity).
Please move this to your CSS file.

css/bodystyle.css
```diff
@@ -0,0 +1,200 @@
...
+
+img.logo {
+  width: 20px;
+  height: auto;
+  margin: 10px;
+  margin-left: 40px;
+}
+
+nav ul {
+  float: right;
```
LucyMac, line 27: Please don't use `float` rules in your CSS. Float is the old way of making layouts in CSS and it has now been replaced by Flexbox (which was introduced a few years ago). 
Set `display: flex` on the parent of the element instead.

css/bodystyle.css
```diff
@@ -0,0 +1,200 @@
...
+  margin: 0 20px;
+}
+
+nav ul li {
+  display: inline;
```
LucyMac, line 32: You can use flexbox on the `ul` instead of inline here.

css/bodystyle.css
```diff
@@ -0,0 +1,200 @@
...
+
+nav ul li a {
+  color: rgb(68, 67, 67);
+  font-size: 20px;
+  padding: 5px 10px;
+}
+
+a:hover {
+  color: rgb(197, 58, 58);
+  transition: 0.5s;
```
LucyMac, line 45: Nice

css/bodystyle.css
```diff
@@ -0,0 +1,200 @@
...
+  background-position: right;
+  color: whitesmoke;
+
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  width: auto;
+  text-align: center;
+  padding: 20%;
```
LucyMac, line 79: You need to add a width here for your image to cover the page.
Try `100%`

css/bodystyle.css
```diff
@@ -0,0 +1,200 @@
...
+  font-size: 15px;
+  margin: 10px;
+}
+
+#hero button {
+  margin-top: 20px;
+  width: 120px;
+  height: 40px;
+  font-size: 14px;
+  background-color: rgb(225, 106, 66);
```
LucyMac, line 98: Can you add a border-radius on the button and remove the default border to match the design?

css/bodystyle.css
```diff
@@ -0,0 +1,200 @@
...
+  ul {
+    position: fixed;
+    width: 50%;
+    height: auto;
+    background: whitesmoke;
+    top: 60px;
+    right: -100%;
+    text-align: center;
+    transition: all 0.5s;
+  }
```
LucyMac, line 185: Love your mobile nav!
Perhaps give this `ul` a class to be a little more precise, and add your CSS to that instead of `ul` (a bit too generic).

css/dummyStyles.css
```diff
@@ -0,0 +1,51 @@
+/* HEADER */
```
LucyMac, line 1: It would be best practice not to include this file if you are not using it

LucyMac (commented):
Good HTML structure and nice mobile version Mustafa 🙂 

Please fix your hero section on desktop - and remember not to use IDs in CSS!

PR #150, opened 2020-12-14
London Class 7 - Soraiya Muhammad - HTML/CSS - Week 2

soraiya:
Name: Soraiya Muhammad
City: London
Slack Name: soraiya

Module: HTML & CSS
Week: 1


css/style.css
```diff
@@ -16,4 +17,227 @@ body {
...
+  background-color: white;
+  align-items: flex-end;
+}
+
+.twitter,
+.fb,
+.insta {
+  width: 30px;
+  height: 30px;
+}
```
LucyMac, line 197: Same here, if you're styling them all the same, you only need one class

css/style.css
```diff
@@ -16,4 +17,227 @@ body {
...
+.nav-links li {
+  display: inline-block;
+  padding: 0px, 20px;
+  margin-left: 30px;
+}
+
+.nav-links li a:hover {
+  color: rgb(199, 82, 40);
+  cursor: pointer;
+  transition: all 0.3s ease 0s;
```
LucyMac, line 48: Try adding `text-decoration: none` to your links to remove the underline.

LucyMac (commented):
Nice work Soraiya, your page looks great on desktop - can you add some media queries to make it work on mobile? 
Remember to try and re-use CSS classes when you can 🙂

PR #174, opened 2021-05-24
Nw-class-4/mya skylar

myaskylar:

# Your Details

- Your Name: Mya Skylar
- Your City: North-west
- Your Slack Name: Mya skylar

# Homework Details

- Module: course-work-wk2
- Week: Week 3

# Notes

- What did you find easy?
  HTML structure.
- What did you find hard?
   To create responsive nav-bar
   Need more practice on media queries.
- Any other notes?
  I've notice to create the hamburger menu bar on mobile screen size needs js knowledge.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -149,6 +149,38 @@ footer p:last-child {
   color: gray;
 }
 
+/*@media*/
+
+@media screen and (max-width: 360px) {
+  header {
+    padding: 0px 10px;
+  }
+
+  .nav-bar > ul {
+    flex-direction: column;
+  }
+
+  h1 :first-child {
+    font-size: 3em;
+  }
+
+  .intro-karma > p {
+    font-size: 1.5em;
+    font-weight: 200;
+    margin-bottom: 50px;
+    text-align: center;
+  }
+
+  h1 {
+    font-size: 2em;
+  }
+
+  .section-container {
+    flex-direction: column;
+    align-items: center;
+  }
+}
+
 /*footer*/
 
 /*For own reference*/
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -16,4 +15,307 @@ body {
...
+}
+
+#toggle {
+  display: none;
+}
+
+/*styling article*/
+
+.intro-karma {
+  background-image: url(/img/first-background.jpg);
```
KarenPudner, line 69: The path to the image is not correct, so it is not displaying. Your CSS file is in a subfolder, so you need to go up a level before going into the img folder

index.html
```diff
@@ -41,6 +41,21 @@ <h1>Introducing Karma</h1>
         <p>Bring WiFi with you, everywhere you go</p>
         <button>Learn More</button>
       </article>
+      <h1>Everyone needs a little Karma.</h1>
+      <section class="device">
```
KarenPudner: You don't seem to refer to the classes "device", "boost" or "pay" in your css file.

KarenPudner (approved):
Good work! The main issue is that the main image is not displaying - I've downloaded it locally, so I can see that it will display correctly if you just tweak the path. You generally target classes rather than IDs in your css, which is good. You also make good use of semantic HTML.  The images seem to respond well to changing the window size. It would be useful if you could include some screenshots of your webpage in different window sizes, and also run Lighthouse against it and include a screenshot in your solution. Let me know if you don't know how to do that.

css/style.css
myaskylar: Thanks for the hint. I've correct the files accessibility.

index.html
myaskylar: I've changed all section class to card and use it in .card CSS to style.

PR #176, opened 2021-05-25
Northwest class 4- Anza Azam - HTML/CSS - Week 2   branch northwest class 4/anza azam 

Anza-Azam:

# Your Details

- Your Name: Anza Azam
- Your City: Northwest
- Your Slack Name: Anza-Azam

# Homework Details

- Module: HTML/CSS
- Week:2

# Notes

- What did you find easy?
HTML semantics
- What did you find hard?
CSS styling
Images path from out of scope in a cloned repo -one image I had to change location
- Any other notes?
media query is still for mobile design as designed desktop-first


index.html
```diff
@@ -59,21 +60,33 @@ <h2>Pay as You go</h2>
           </article>
         </div>
       </section>
+      <section id="extended">
+
+          <img id="feature" src="/img/feature-img.png" alt="">
```
KarenPudner, line 66: Ideally use vector images like .svg (which you have elsewhere) rather than raster images - more information here https://www.shutterstock.com/blog/raster-vs-vector-file-formats

index.html
```diff
@@ -1,19 +1,98 @@
...
+         
+          <ul id="list">
+            <li><a href="">Meet Karma</a></li>
+            <li><a href="">How it Works</a></li>
+            <li><a href="">Store</a></li>
+            <li><a href="">Blog</a></li>
+            <li><a href="">Help</a></li>
+            <li><a href="">Login</a></li>
+          </ul>
+           <img id="hamburger" src="img/menu-hamburger.svg" alt="" />
```
KarenPudner, line 34: You have correctly added alt text to most images, but have missed a couple.

KarenPudner (approved):
Generally very good. You have an excellent Lighthouse score, you make use of semantic HTML,  you use flex effectively and the resulting webpage looks very similar to the model answer. A couple of points for improvement - it is good practice to target classes rather than ID in CSS. Even if the class currently only applies to 1 element, it makes it easier to extend the style if you add similar elements later. Also, the button does not have the same background as the model answer - was this intentional and due to wanting to get a higher Lighthouse score? And a couple of images to not have alt text. Also, a few details like text style and the Facebook/Twitter/Instagram icons also look different to the model answer. But overall, very good.

Anza-Azam:
Thank you Karen, I'm going to do the changes as you have mentioned for improvement classes, alt, svg, yes the background color was for the lighthouse point of view, I wanted a 100 score, will add every other change and increase the match to match more, add alt and try using svg images. I really appreciate your review.

PR #180, opened 2021-05-28
NW class Terry Calderbank HTML/CSS Week 2

Texx45:

# Your Details

- Your Name: Terry Calderbank
- Your City: NW
- Your Slack Name: Terry

# Homework Details

- Module: HTML/CSS Karma site 
- Week: 2

# Notes

- What did you find easy?

Styling with CSS

- What did you find hard?

Positioning everything with flexbox

I am happy with my understanding in general I just think practice is required to solidify. I have learnt a lot from this project as I have had to try many different ways to do something to get it to work.

- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -24,10 +24,8 @@ body {
   align-items: center;
 }
 
-.logo {
-  width: 2rem;
-  padding: 1rem 2.5rem;
-}
+/* .logo {
+} */
 
 .logo img {
   min-width: 40px;
@@ -99,7 +97,7 @@ body {
   justify-content: space-between;
   flex-wrap: wrap;
   width: 80%;
-  margin: auto;
+  margin: 1rem auto 3rem;
 }
 
 .info-icons h4 {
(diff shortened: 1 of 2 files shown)


SteveLeicester:
Great piece of work again Terry

HTML nicely formatted - and good use of semantic HTML.  So the page is nicely divided with a header, main and footer and your main content split into sections.  Plus clear navigation elements too. "Tying" up your info icons together with associated text (as a h4) in  a self contained parent (div) container shows an excellent understanding too.  Nice use of alt text on your images.   Would be tempted to move the size of the icons into your css.  

CSS nicely formatted too.   Good use of various units - rem units in particular.  Great use of flexbox on your navigation  and navbar and your icons too.  Your mobile version looks excellent - switching to column direction for your icons is the way to go on a narrow form factor - nice.  Can you think how to center the icons and associated text?

Suggestion around a good habit to develop is commit and with helpful comments.  Your are starting to do that I can see.  Gets easier if you have a plan.  So in approaching this you can see there is a logo, navigation, header, some icons - probably links of course and a footer.  Make a plan of work around that. So then you might get the navigation in place - commit with comment.  Add the logo - commit,  layout so logo on the left, navigation on the right - commit. And so on.   Always helps if you completely mess things and it's not obvious why you can revert to an earlier commit knowing it's not 4 days ago but only a couple of hours (for example).

Agree with your comment - your understanding is well demonstrated and yes just practice, practice. 

Good stuff
Steve

Texx45:
Thanks for the feedback Steve. I totally agree with your comments about commits, it's a habit I will definitely try to work on as I kept forgetting. I think I need to slow it down a touch and be more methodical and structured like you described. I think I kept getting carried away and then committing as an afterthought. 

This weeks project is another mock up website from a wireframe which I am starting tomorrow. I will aim to stick to the principles we have discussed and work on a more structured approach especially with the commits. 

I will also open my project and  take a look at your suggestion about centring the icons and associated text. 

Thanks for taking the time to review my work and offer feedback it is much appreciated.

Terry

PR #183, opened 2021-05-30
Northwest class 4/konika

konikalily:

# Your Details

- Your Name:Konika Biswas
- Your City: NorthWest

- Your Slack Name:Konika Biswas

# Homework Details

- Module:HTML-CSS-Coursework
- Week:2

# Notes

- What did you find easy?
html., some styling
- What did you find hard?
media queries, grid
making responsive, hamburger menu
- Any other notes?


peedeerich:
Hi Konika,

Well done on your submission! The site is responsive and looks good in terms of layout. Your HTML is mostly semantic with sensible use of elements like <section>, <footer> etc. There are one or two improvements you could make in terms of class names etc which I have highlighted in the comments.

The layout looks good but there are a couple of places where the CSS grid approach used is a little complex - we have covered more on this topic since this homework was set but if you are finding it tricky to translate the desired layout to CSS this is a useful reference 
https://css-tricks.com/snippets/css/complete-guide-grid/

There are some inconsistencies in terms of code formatting both in the HTML and CSS - getting this right really helps make your code more readable and easier to manage and refactor. The good news is you should be able to correct most errors automatically in VS Code :)

Overall a very good effort, well done!

css/style.css
```diff
@@ -16,4 +16,396 @@ body {
...
+        grid-column: 1 / 2;
+        grid-row: 1 / 2;
+    }
+    #get_karma h4{
+        grid-column: 2 / 4;
+        grid-row: 1 / 2;
+        justify-self: center;
+        align-self: center;
+        display: flex;
+        flex-direction: column;
```
peedeerich, line 366: This header element does not contain any children - it is a part of the "get karma" layout and does not control layout itself, so should not be a flexbox. Remember `display: flex` is used on the flex container - it will then change how the child elements are displayed

css/style.css
```diff
@@ -16,4 +16,396 @@ body {
...
+        font-style: italic;
+        font-weight: 400;
+        font-size: 1.5em;
+        color: var(--secondary);
+          
+    }
+    .karma-btn{
+        grid-column: 2 / 4;
+        grid-row:  1;
+        margin-top: 10em;
```
peedeerich, line 378: it seems like this margin is being used to correct an issue in the grid layout - it looks fine but it would be better to work on the layout so that it is not required. This would make it easier to make changes to the layout and dimensions without needing to tweak the margin-top value

hamburger_menu.js
```diff
@@ -0,0 +1,7 @@
+
+var toggleButton = document.getElementsByClassName('toggle-button')[0];
+var navLink = document.getElementsByClassName('nav-link')[0];
+toggleButton.addEventListener('click',showMenu);
+function showMenu(){
+    navLink.classList.toggle("active");
+}
```
peedeerich, line 7: although not in the scope of this homework the hamburger menu using is working well - we will revisit wrking with the DOM in JavaScript later in the JS modules :)

index.html
```diff
@@ -15,5 +15,69 @@
...
+                <h3>Boost your productivity</h3>
+            </article>
+            <article class="service-card">
+                <img src="/img/icon-refill.svg" alt="refilling petrol picture">
+                <h3>Pay as you go</h3>
+            </article>
+            </div>
+        </section>
+
+        <section id="get_karma" class="grid">
```
peedeerich, line 63: Most of your class and id names are sensibly chosen - good work :) Here the class name of "grid" is quite generic, but there are other grids on the page which have quite different properties. In this case it would be better for the class name to be slightly more specific to the feature on the page.

konikalily: Hi Paul, thank you for your suggestion. I made the change to the class name to get_karma-grid. Kind regards.

hamburger_menu.js
konikalily: Thank you for your comment.

css/style.css
konikalily: Hi Paul, Thank you for your comment. I deleted the margin. And added the grid template column in the get karma section container. I hope I have done the right this time. I do struggle with lay out issues. I will practise more and  more . Kind regards.

css/style.css
konikalily: Hi Paul, thank you for pointing this out. I think now I have slightly better understanding of how to use flex box. I made the change now. However I wanted to look the h4 in a certain way in the column in a bigger screen. Couldn't figure out how to do it . Regards.

PR #184, opened 2021-05-30
Liverpool nw4/khmdagal

khmdagal:
<!--

Liverpool NW4 - Khadar M Dagal - HTML-CSS-Coursework-Week2

Please complete the details below this message

# Notes

- What did you find easy?
 Generally it was ok

- What did you find hard?
what it found out really difficult was how to connect the 'toggle' and the checkbox  without JS only CSS, so they work together while the checkbox is hidden.
I have use Display:none; and visibility:hidden; both did not work.

 definitely I need more practice on grid, and to understand more how they work flex in different purpose 

- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,19 +1,210 @@
 
- /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
 
-body {
-    font-family: 'Roboto', sans-serif;
-    -webkit-font-smoothing: antialiased;
-}
+/* phone size quary */
 
-/**
- * Add your custom styles below
- * 
- * Remember: 
- * - Be organised, use comments and separate your styles into meaningful chunks
- *    for example: General styles, Navigation styles, Hero styles, Footer etc.
- * 
- * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
- */
+@media only screen and (max-width: 600px){ 
+    .body {
+        font-family: 'Roboto', sans-serif;
+        -webkit-font-smoothing: antialiased;
+        display: inline-block;
+        grid-template-rows: auto;
+        background-color: rgb(27, 196, 139);
+        }
+        
+        label {
+            display: block;
+            cursor: pointer;
+            width: 30px;
+            font-size: 20px;
+            float: right;
+            
+        }
+        .navigation{
+            display: grid;
+            grid-template-columns: 90% 10%;
+        }
 
+        .Nav a:hover, button:hover {
+            color: rgb(59, 25, 12);
+            font-style: italic;
+            font-weight: bolder;
+        }
+        #toggle{
+            visibility: hidden;
+        }
 
+        #toggle, .toggle{
+            grid-column: 2;
+            grid-row: 1;
+            position:relative;
+            
+        }
+        .Nav {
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -1,19 +1,210 @@
...
+    .footer {
+        display: flexbox;
+        text-align: center;
+    }
+    
+    
+    .footer-icons {
+        width: 8%;
+        height: 8%px;
+        justify-content:space-between;
```
zeinaji, line 207: You either want 8px or 8% here I think :)

css/style.css
```diff
@@ -1,19 +1,210 @@
...
+  @media screen and (min-width:601px) and (max-width: 2000px) {
+
+    .body {
+        font-family: 'Roboto', sans-serif;
+        -webkit-font-smoothing: antialiased;
+        display: grid;
+        grid-template-columns: auto;
+        background-color: rgb(140, 240, 140);
+        }
+
```
zeinaji, line 121: Nice work with making it all responsive! 

It's worth noting though that instead of having to repeat the same code for each `@media` block, you can have the stuff common for both designs outside of the `@media` blocks. This way `@media` will only be used for things that should be different on different sized screens

css/style.css
```diff
@@ -1,19 +1,210 @@
...
+        flex-direction: column;
+    }
+    .maintitle {
+       position: relative;
+       top: -100px;
+       text-align: center;
+       color: lavenderblush;
+       font-size: 20px;
+        }
+ .main {
```
zeinaji, line 80: Just be careful about spaces between rules and general indentation. All the selectors should be aligned here :)

index.html
```diff
@@ -6,14 +6,64 @@
...
     <link rel="shortcut icon" type="image/x-icon"  href="favicon.ico">
 </head>
-<body>
+<body class="body">
+    <header class="header">
+     <nav class="navigation">
+        <img class="logo" src="img\karma-logo.svg" alt="Karma Logo">
+        <label class="toggle">&#9776;</label>
+        <input type="checkbox" id="toggle">
+
```
zeinaji, line 18: Nice work with the hamburger menu on smaller screens 🙂 At the moment though when I click on it I'm not getting any links

index.html
```diff
@@ -6,14 +6,64 @@
...
+            <img src="img\icon-refill.svg"  class="icons" alt="fuel tang icon">
+            <h3>Pay as You Go</h3>
+        </section>&nbsp;  
+    </main>
+
+    <footer class="footer">
+        <h5>Join us on</h5>
+        <img  class="footer-icons" src="img\twitter-icon.svg" alt="Twitter Logo">&nbsp;&nbsp;
+        <img class="footer-icons" src="img\facebook-icon.svg" alt="Facebook Logo">&nbsp;&nbsp;
+        <img class="footer-icons" src="img\instagram-icon.svg" alt="Instagram Logo">&nbsp;&nbsp;
```
zeinaji, line 64: It's worth adding an empty link to these, just to give the impression that you can click on them when hovering over 🙂

zeinaji (commented):
Nice work with semantic HTML and flexbox 🙂

PR #186, opened 2021-05-31
NW4 - Azan Shaikh - HTML/CSS - Week 2

Ash-97:

# Your Details

- Your Name: Azan Shaikh
- Your City: NorthWest 4
- Your Slack Name: Azan

# Homework Details

- Module: HTML/CSS
- Week: 2

# Notes

- What did you find easy?
Adding all the required html and styling it.
- What did you find hard?
It wasn't hard per say but took a decent bit of time to make the page responsive. 
How to make a dropdown menu instead of a standard navbar. 
- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -28,6 +28,10 @@ header img {
   padding: 5px;
 }
 
+.dropdown-menu {
+  display: none;
+}
+
 .navbar ul {
   margin: 0 7% 0 0;
   display: flex;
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -58,10 +62,11 @@ header img {
 .intro {
   background-image: url(/img/first-background.jpg);
```
KarenPudner, line 63: The image is not displaying. The css file is in a subfolder so your path needs to take this into account.

KarenPudner (approved):
Good work! The webpage looks correct apart from the broken image link. You are correctly targeting classes rather than IDs in your css, and the classes are well named. Good use of flex.

css/style.css
Ash-97: > The image is not displaying. The css file is in a subfolder so your path needs to take this into account.

Hi Karen, I am not sure what the issue is because on my laptop the image is only being displayed if I set it with the path that I have already used.

PR #188, opened 2021-06-02
Manchester NW1 - Erin_Dyson - HTML/CSS - Week 3

erin-switchstitch:

# Your Details

- Your Name: Erin Dyson
- Your City: Manchester
- Your Slack Name: erin-switchstitch

# Homework Details

- Module: HTML / CSS
- Week: 3

# Notes

- What did you find easy?

With the tutorial posted on the slack channel, I found creating the nav bar using a checkbox much easier than navbars I've made before with Javascript

- What did you find hard?

I found working out how to set e.g. the main three icons (coffee, refill, devices) pretty difficult to resize so that they are all the same size. I could really use some help with the theory behind how to target these icons and change their height / width and when to use fixed values and when to use relative values. Also what to do if e.g. two of the icons are the right size, and one of them is smaller than the others .... how do we find out why the third icon is smaller.

I could really use doing some more work on when we should use flex, flexbox, inline-flex etc... Also could use some help knowing the difference between justify items, self and content - and align content, self, items, text ....

- Any other notes?


index.html
```diff
@@ -5,15 +5,91 @@
...
+            <a class="nav-a">Help</a>
+            <a class="nav-a">Login</a>
+        </div>
 
-    <!-- Add your HTML markup here -->
-    <!-- Remember: Use semantic HTML tags like <header>, <main>, <nav>, <footer>, <section> etc -->
-    <!-- All the images you need are in the 'img' folder -->
+    </nav>
+
+    <article>
```
peedeerich, line 31: Semantically, the header is a "top level" element in your document and so it doesn't make sense for it to be wrapped with <article> here - removing it shouldn't alter the layout

index.html
```diff
@@ -5,15 +5,91 @@
...
+        <header> 
+            <div class="header-div">
+                <div class="header-image">  </div>
```
peedeerich, line 34: I think (though haven't tried) you should be able to get the background image showing by including it in the CSS of "header-div" or "header" - this would be better than it having its own element

index.html
```diff
@@ -5,15 +5,91 @@
...
+                
+                <!-- <img class="header-image" src="img\first-background.jpg" alt="An ipad and karma Wifi device on a table with a person in a white shirt looking at these devices"> -->
```
peedeerich, line 36: Using CSS to bring in the background image as you have done seems the best thing to do here but it raises an interesting question around accessibility - what about alt text?

 where background images are purely decorative the general rule is that they won't need alt text because they don't convey information to the user.

index.html
```diff
@@ -5,15 +5,91 @@
...
+        <p>Join us on</p>
+        
+        <div class="footer-container">
+            
+            <i class="fab fa-twitter"></i>
+            <i class="fab fa-facebook-f"></i>
+            <i class="fab fa-instagram"></i>
+
+        </div>
+    </footer>
```
peedeerich, line 92: the footer should be inside the body tag (anything that forms part of the document structure goes in body). Browsers are sometimes a bit too helpful - most will fix this when you open the page! Great for a user but a bit annoying when you are developing :)

index.html
```diff
@@ -5,15 +5,91 @@
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     <title>Karma</title>
     <link href='https://fonts.googleapis.com/css?family=Roboto:400,500,300' rel='stylesheet' type='text/css'>
+    <script src="https://kit.fontawesome.com/c5012398d3.js" crossorigin="anonymous"></script>
+    
     <link rel="stylesheet" href="css/normalize.css">
     <link rel="stylesheet" href="css/style.css">
     <link rel="shortcut icon" type="image/x-icon"  href="favicon.ico">
+    <link rel="stylesheet" type="text/css" media="screen and (min-device-width: 800px)" href="css/over800.css" />
```
peedeerich, line 13: Using different stylesheets with media queries is perfectly valid but you need to take care that only one sheet is being applied at a time - in this case when you are on a wider device both will be active (I think you have noticed this based on comments). This is true of media queries in general but tends to cause more of a problem with this structure as it affects the whole stylesheet rather than a few selectors

peedeerich:
Hi Erin - well done on this project! The site looks very good - there are some slight issues with the nav bar on smaller screen sizes (I am using Chrome) but nothing major. 

I have left some specific comments - also as with the previous homework there are some improvements you could make in terms of code formatting (with the same advice - let VSCode sort it!)

PR #195, opened 2021-06-04
North west 4/maziar majd

Maziarmajd:

# Your Details

- Your Name: Maziar Zahedi Majd
- Your City: Bolton
- Your Slack Name: maziarmajd

# Homework Details

- Module: html/css
- Week:2

# Notes

- What did you find easy?
media query
- What did you find hard?
grid, 
grid and flexbox
- Any other notes?


SteveLeicester:
Thankyou for submitting. This is a great piece of work Maziar.

Good use of GitHub - plenty of small commits is the way to do it.

Your HTML is clear and nicely formatted.  Good use of semantic HTML with a header, main and footer.  The header as the logo and navigation - nicely done with list.


You main content is divided into three pieces - with a background image and titles, followed by the icons and then well done for adding the extended content.

I would suggest these three could be sections. rather than div.  Also try to have heading in order  h1 then h2 rather than h1 h3.  You can change the style of the heading if it doesn't look right.

Nicely formatted CSS with a mix of different units. Well done for the use of flexbox and grid.  Good to see the navigation list as a flex horizontal (the default).

You have used a number of id selectors for example #title, #subtitle, #icons.  These are intended to style one unique element (and can be used for internal links) - so in your HTML you have repeated img elements with id="icons".  id should be unique so you should be using class selectors.
You can read more about that here

[https://css-tricks.com/the-difference-between-id-and-class/](https://css-tricks.com/the-difference-between-id-and-class/)


Your responsive design works well but there are improvements. Mobile looks good (though the extended content disappears). But resizing your icons shrink. They look a little small if you shrink a browser window.  And your headers collide. 

Nevertheless you should be very pleased with this work.  

Yes - appreciate your comment - grid and flexbox are not easy to understand.  They take practice but remember if you aim to build simple elegant web pages then many of the complex features of flex and grid are not needed.

Keep it up.


Steve

PR #216, opened 2021-06-10
Scotland class 5 - Ivan Zhukov - HTML/CSS - Week 2

zhkvivan:

# Your Details

- Your Name: Ivan Zhukov
- Your City: Glasgow
- Your Slack Name: Ivan Zhukov

# Homework Details

- Module: HTML/CSS
- Week: 2

# Notes

- What did you find easy?
flexbox
- What did you find hard?
mobile burger menu

- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -74,6 +74,15 @@ header .container {
 	width: 20px;
 }
 
+.burger-menu__icon {
+	width: 20px;
+	height: 2px;
+	background-color: black;
+	margin: 5px 0;
+}
+
+
+
 h1 {
 	font-weight: 200;
 	margin-bottom: 0px;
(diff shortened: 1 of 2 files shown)


sriram-siv:
This is great work Ivan, well done :)

The css solution for the mobile menu is really cool, not something I've seen before.
Very minor issue would be nesting a <nav> inside another <nav>. Generally you want to avoid this with the semantic elements and just use a div for the inner container.
Regarding the commit messages, I would say they are mostly fine. Some are a bit generic ('adding some text') so could be a bit more descriptive. I think that it sometimes helps to plan ahead what work you are doing so that when you commit it will be grouped together in a meaningful way, but this will make more sense once you're working on a larger project.

PR #220, opened 2021-06-11
Glasgow Class 5 - Andrew Robertson - HTML/CSS - Week 2

Andy-Robertson:
# Your Details

- Your Name: Andrew Robertson
- Your City: Glasgow
- Your Slack Name: Andy

# Homework Details

- Module: HTML/CSS
- Week: 2

# Notes

- What did you find hard?
I found it hard to say DONE and submit as I kept asking myself if I just made this change then it would look even closer to the design pics (which happened a lot).


css/style.css
```diff
@@ -1,19 +1,506 @@
+body {
+  font-family: "Roboto", sans-serif;
+  -webkit-font-smoothing: antialiased;
+}
 
- /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
+/* Header */
+.header-section {
+  display: flex;
+  flex-flow: row nowrap;
```
bonboh, line 9:
```suggestion
```
`flex-direction` is `row` by default, and `flex-wrap` is `nowrap` by default, so we don't need this

css/style.css
```diff
@@ -1,19 +1,506 @@
...
+  width: 100%;
```
bonboh, line 10:
```suggestion
```
The width is 100% already, so we don't need this. Same applies to other code in this file where width is already the maximum size

css/style.css
```diff
@@ -1,19 +1,506 @@
...
+  padding: 1rem 0 1rem 0;
```
bonboh, line 11:
```suggestion
  padding: 1rem 0;
```
The shorthand is typically used, so we can do the same. Same goes for the rest of this file

css/style.css
```diff
@@ -1,19 +1,506 @@
...
+.navbar {
+  display: flex;
+  flex-flow: row wrap;
+  justify-content: flex-end;
+  width: 100%;
+}
+
+.navbar-hamburger-Menu {
+  display: flex;
+  flex-flow: column nowrap;
```
bonboh, line 36:
```suggestion
```
This button element has one child which is an image. Creating a flex column doesn't do anything here, so we should remove it

css/style.css
```diff
@@ -1,19 +1,506 @@
...
+  font-weight: 500;
+}
+
+.navbar-full-Menu > li > a {
+  line-height: 3rem;
+  color: #89899e;
+  text-decoration: none;
+}
+
+.navbar-full-Menu > li > a:hover {
```
bonboh, line 68:
```suggestion
.navbar-full-Menu > li > a:focus,
.navbar-full-Menu > li > a:hover {
```
When adding hover, we should typically add focus too so keyboard users pressing the `Tab` key to jump to the button get this style too

css/style.css
```diff
@@ -1,19 +1,506 @@
...
+  color: #e0633a;
+  cursor: pointer;
+}
+
+/* Hero */
+.hero-section {
+  display: flex;
+  flex-flow: column nowrap;
```
bonboh, line 76:
```suggestion
  flex-direction: column;
```
`flex-wrap` is `nowrap` by default, so we don't need to specify `nowrap`. Same goes for the other instances in this file

css/style.css
```diff
@@ -1,19 +1,506 @@
...
+  padding-top: 2rem;
+}
+
+.hero-section > button {
+  background-color: #e0633a;
+  border: 0px;
+  border-radius: 4px;
+  margin-top: 2rem;
+  width: 7.4rem;
+  height: 2.8rem;
```
bonboh, line 105:
```suggestion
  padding: 1rem;
```
Rather than set a specific width and height, we can use padding. The benefit is that the width sizes nicely to the text content whether there's a little or a lot of text.

css/style.css
```diff
@@ -1,19 +1,506 @@
...
+}
+
+.hero-section > button:hover {
+  background-color: #b15030;
+  transition: background-color 0.8s ease-in;
```
bonboh, line 110:
```suggestion
  transition: background-color 0.1s ease-in;
```
Transitions like this are typically short and snappy, so we could change this to something like `0.1s`

css/style.css
```diff
@@ -1,19 +1,506 @@
...
+.about-icon {
+  width: 5.6rem;
+}
+
+.about-section {
+  display: flex;
+  flex-flow: column wrap;
+  align-items: center;
+  width: 100%;
+  margin: 0;
```
bonboh, line 127:
```suggestion
```
It looks like the margin is already 0 so we don't need this

index.html
```diff
@@ -1,19 +1,122 @@
...
+        <button>Learn More</button>
+      </section>
+      <section class="about-section">
+        <h2>Everyone needs a little Karma.</h2>
+        <article class="about-articles">
+          <div class="about-container">
+            <img
+              class="about-icon"
+              src="img/icon-devices.svg"
+              alt="Picture of a smartphone, table and desktop computer"
```
bonboh, line 61:
```suggestion
              alt="A smartphone, table and desktop computer"
```
We don't need to mention "picture" in the description as we already know it's a picture. Same applies to the other alt descriptions

index.html
```diff
@@ -1,19 +1,122 @@
...
+            />
+            <h3>Internet for all devices</h3>
```
bonboh, line 63:
```suggestion
            <p>Internet for all devices</p>
```
This doesn't seem like a heading of a section, so it seems more appropriate as a `p` tag

index.html
```diff
@@ -1,19 +1,122 @@
...
+              src="img/icon-refill.svg"
+              alt="Pay as You Go"
+            />
+            <h3>Pay as You Go</h3>
+          </div>
+        </article>
+      </section>
+    </main>
+    <footer class="footer">
+      <h4>Join us on</h4>
```
bonboh, line 86:
```suggestion
      <h2>Join us on</h2>
```
This seems like it can be an `h2`, as it's not a subsection of the "Everyone needs a little Karma." `h2` tag

bonboh (commented):
Great job @Andy-Robertson! It looks great on all screen sizes 💯. The main things to note are that we don't need to add code when it's already a default, and when it doesn't do anything. I've left some comments and tips for you to read, but you don't need to do anything else 👍

css/style.css
Andy-Robertson: That's awesome, I didn't think about doing shorthand on shorthand.

css/style.css
Andy-Robertson: This is an oversight, I realise its good practice, thanks for pointing it out :+1:

css/style.css
Andy-Robertson: This is something I will take forward into the week 3 project, thanks!

Andy-Robertson:
Thank you for taking the time to providing feedback, its very much appropriated and I will try to be mindful of it in future projects.

PR #227, opened 2021-06-14
Scotland class 5 - karma clone

bonboh:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -54,7 +54,8 @@ body {
               padding:8px; 
              }
     
-   /*.nav-link:first-child {color: rgb(46, 44, 44)}*/
+   .nav li:first-child > a {color: rgb(46, 44, 44);}
+
 
 article {margin-top:70px;}
 
@@ -66,13 +67,18 @@ article {margin-top:70px;}
                   height: 700px;
                   width:100%;
                   background-size: cover left;
-                 background-position: center bottom;
-                 padding-top:50px;
-                 text-align: center;
-                 background-repeat: no-repeat;
-                  } 
-   #into-content {align-items: center;}
-                
+                  background-position: center;
+                  padding-top:50px;
+                  text-align: center;
+                  background-repeat: no-repeat;
+                        
+                  display: flex;
+                  justify-content: center;
+                  align-items: center;
+                   } 
+
+   #into-content {}
+
 #learnbutton { background-color: #F15B2A;
         border: none;
         color: white;
--- a/css/style.css
+++ b/css/style.css
@@ -4,7 +4,12 @@
 body {
     font-family: 'Roboto', sans-serif;
     -webkit-font-smoothing: antialiased;
-}
+    width:100%;
+    height:100%;
+    text-align: center;
+    line-height: 25px;
+    box-sizing: border-box;
+    }
 
 /**
  * Add your custom styles below
@@ -16,4 +21,79 @@ body {
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
 
+#nav-bar {position:fixed; 
+          top:0; 
+          height: 70px;
+          width:100%;
+         }
 
+    #nav-bar ul {list-style-type:none; 
+                display:flex; 
+                justify-content: flex-end;
+                }     
+  
+    #nav-bar ul a:hover {color:#F15B2A;
+                        }
+
+    #nav-bar img {height:60px; 
+                  width:200px; position: absolute;
+                  left: 0px;
+                  top: 10px;
+                  z-index: -1;
+                 }
+
+    .nav-link {text-decoration:none; 
+              color:darkgray; 
+              font-weight:bolder; 
+              display:block; 
+              padding:8px; 
+             }
+    
+   /* .nav-link:first-child {color: rgb(46, 44, 44)}*/
+
+article {margin-top:70px;}
+
+
+   h1, h2 {color: white; font-weight:lighter;}
+
+#intro {background-image: url('https://raw.githubusercontent.com/willem-blauenberg/portfolio/master/projects/karma/img/first-background.jpg'); 
+                  height: 700px;
+                  width:100%;
+                  background-size: cover left;
+                 background-position: center bottom;
+                 text-align: center;
+                 background-repeat: no-repeat;
+                 margin:1px;
+                                  } 
+                    
+                
+button { background-color: #F15B2A;
+        border: none;
+        color: white;
+        padding: 15px 32px;
+        text-align: center;
+        text-decoration: none;
+        display: inline-block;
+       }
+
+#content {height:30%}
+
+#content h1 {color: rgb(46, 44, 44);}
+
+.service-img {width:100px; height:100px; }
+
+.services {width:200px;
+           text-align:center;
+           display:inline-block;
+           float:center;
+           padding:10px;
+           }
+
+.servicename {text-decoration: none; color:rgb(46, 44, 44);}
+
+hr {width: 75%; color:lightgrey; align-items: center;}
+
+footer {height:20%;}
+
+
+  
--- a/css/style.css
+++ b/css/style.css
@@ -3,12 +3,15 @@
 
 body {
     font-family: 'Roboto', sans-serif;
+    font-size: 16px;
     -webkit-font-smoothing: antialiased;
     width:100%;
     height:100%;
+    color:#838994;
     text-align: center;
     line-height: 25px;
-    box-sizing: border-box;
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -3,8 +3,16 @@
 
 body {
     font-family: 'Roboto', sans-serif;
+    font-size: 16px;
```
bonboh, line 6:
```suggestion
    font-size: 1rem;
```
We should use relative units like `rem`, especially for `font-size`. If we use pixels, then if a user changes their browser default font size setting then the screen doesn't increase in size

css/style.css
```diff
@@ -3,8 +3,16 @@
...
     -webkit-font-smoothing: antialiased;
-}
+    width:100%;
+    height:100%;
```
bonboh, line 9:
```suggestion
```
We don't need these properties so we can remove them

css/style.css
```diff
@@ -3,8 +3,16 @@
...
+    color:#838994;
+    text-align: center;
+    line-height: 25px;
+    box-sizing: border-box; /*padding included in the box size*/
+    display: block; /*each element as a block*/
```
bonboh, line 14:
```suggestion
```
Normally, it's not a good idea to turn every element into a block, because when we use inline elements like `<a>` and `<strong>` for example, they will all be on separate lines which does not make sense

css/style.css
```diff
@@ -16,4 +24,101 @@ body {
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
 
+#nav-bar {position:fixed;
+          top:0; 
+          height: 70px;
```
bonboh, line 29:
```suggestion
          height: 4.5rem;
```
We should use relative units where possible. Same applies to all the other uses of `px` in this file

css/style.css
```diff
@@ -16,4 +24,101 @@ body {
...
+          background-color: white;
+                   }
 
+    #nav-bar ul {list-style-type:none; 
+                display:flex; 
+                justify-content: flex-end;
+                }     
+  
+    #nav-bar ul a:hover {color:#F15B2A;
+                        }
```
bonboh, line 41:
```suggestion
#nav-bar ul a:focus,
#nav-bar ul a:hover {
  color:#F15B2A;
}
```
We should also add `:focus` so that keyboard users who use the `Tab` key to jump to the link also get a change in colour.

Also, for coding the formatting of the code matters in software teams. Typically, it's structured like in my suggestion. This applies to the whole file

css/style.css
```diff
@@ -16,4 +24,101 @@ body {
...
+
+    #nav-bar img {height:60px; 
+                  width:200px; position: absolute;
+                  left: 10px;
+                  top: 10px;
+                  z-index: -1;
```
bonboh, line 47:
```suggestion
```
`z-index` is not needed here so we can remove it

css/style.css
```diff
@@ -16,4 +24,101 @@ body {
...
+             }
+    
+   .nav li:first-child > a {color: rgb(46, 44, 44);}
+
+
+article {margin-top:70px;}
+
+
+   h1 {color: white; font-weight:lighter; font-size: 2em;}
+   h2 {color: white; font-weight:lighter;}
```
bonboh, line 64:
```suggestion
   #intro h2 {color: white; font-weight:lighter;}
```
We want the `h2` to only be white in the `#intro` section, so we can update the selector to only apply to the `h2`s in the intro

css/style.css
```diff
@@ -16,4 +24,101 @@ body {
...
+
+#intro {background-image: url('https://raw.githubusercontent.com/willem-blauenberg/portfolio/master/projects/karma/img/first-background.jpg');
```
bonboh, line 66:
```suggestion
#intro {background-image: url('/img/first-background.jpg'); 
```
There's an image in the codebase that we can use, which is better than linking to an external image

css/style.css
```diff
@@ -16,4 +24,101 @@ body {
...
+                  padding-top:50px;
+                  text-align: center;
+                  background-repeat: no-repeat;
+                        
+                  display: flex;
+                  justify-content: center;
+                  align-items: center;
+                   } 
+
+   #into-content {}
```
bonboh, line 80:
```suggestion
```
As this rule is empty, we should delete it because software teams don't leave empty code in the codebase

css/style.css
```diff
@@ -16,4 +24,101 @@ body {
...
+#content {height:350px}
+
+#content h1 {color: rgb(46, 44, 44); }
+
+.service-img {width:100px; height:100px; }
+
+.services {width:200px;
+           text-align:center;
+           display:inline-block;
+           float:center;
```
bonboh, line 100:
```suggestion
```
`float` doesn't apply here so we can remove it

css/style.css
```diff
@@ -16,4 +24,101 @@ body {
...
+           padding:10px;
+           }
+
+.servicename {text-decoration: none; color:rgb(46, 44, 44);}
+
+
+footer {height:220;
```
bonboh, line 107:
```suggestion
footer {height:220px;
```
It looks like we're missing units here. Also, we should be using relative units like `rem` as mentioned in my other comment

css/style.css
```diff
@@ -16,4 +24,101 @@ body {
...
+
+footer li {padding:10px;
+           display:inline-block;
+          }
+
+footer ul li img { width: 30px;
+          } 
+
+footer ul li img : hover {cursor:pointer;
+                         }
```
bonboh, line 123:
```suggestion
```
The icons aren't clickable so we shouldn't add a mouse pointer on hover. If we want a mouse pointer on hover, we should put the icons in a link `<a>`

index.html
```diff
@@ -9,11 +9,71 @@
...
+       <li> <a class="nav-link" href="#Howitworks"> How it works </a> </li>   
+       <li> <a class="nav-link" href="#Store"> Store </a> </li> 
+       <li> <a class="nav-link" href="#Blog"> Blog </a> </li>
+       <li> <a class="nav-link" href="#Help"> Help </a></li>
+       <li> <a class="nav-link" href="#Login"> Login</a></li> 
+     </ul>
+  </nav> 
+</header> <!--header ends here-->
+
+<article>
```
bonboh, line 34:
```suggestion
<main>
```
This seems like it should be a main, since we normally have `header` `main` `footer` as the top level elements on a page

index.html
```diff
@@ -9,11 +9,71 @@
...
+
 <body>
 
     <!-- Add your HTML markup here -->
     <!-- Remember: Use semantic HTML tags like <header>, <main>, <nav>, <footer>, <section> etc -->
     <!-- All the images you need are in the 'img' folder -->
 
+    <!-- header starts here-->
+<header id="header">  
+   <nav id="nav-bar">
```
bonboh, line 21:
```suggestion
<header class="header">  
   <nav class="nav-bar"> 
```
Normally, software teams use classes for styling and not IDs. The same applies for the rest of this file

index.html
```diff
@@ -9,11 +9,71 @@
...
+       <h1> Introducing Karma </h1>
+       <h2> Bring WiFi with you, everywhere you go.</h2>
+       <button type="button" id="learnbutton"> Learn More </button>
+     </div>
+   </section>
+   <!-- Introduction ends here-->
+
+   <!--content starts here-->
+   <section id="content"> 
+      <h1> Everyone needs a little Karma.</h1>
```
bonboh, line 47:
```suggestion
      <h2> Everyone needs a little Karma.</h2> 
```
We should only have one `h1` on the page, so we should update this to `h2`

index.html
```diff
@@ -9,11 +9,71 @@
...
+        <img class="service-img" src="img\icon-refill.svg" alt="Pay as You Go"/>
+        <p class="services"> <a href = "" target="_blank" class="servicename" > Pay as You Go </a> </p>
+       </div>        
+   </section>
+   <!-- content ends here-->
+</article>
+
+
+<footer>
+ <p> Join us on </p>
```
bonboh, line 69:
```suggestion
 <h2> Join us on </h2>
```
This seems like it could be a heading, so we could make it one

bonboh (commented):
Great job @M-hayek! Nice to see semantic HTML 👍 I've left some comments and tips for you to read, but you don't have to do anything else. Generally, the main things to note are:

- Software teams typically use class names for styling, not ID, so the HTML document doesn't need to have as many IDs
- Code formatting matters to software teams, so try and style it with proper indents and style
- We should use relative units like `rem`s by default rather than `px`s

PR #230, opened 2021-06-15
Scotland class 5 - Tara Quinn - HTML/CSS - Week Two

Tara-Q:

# Your Details

- Your Name: Tara
- Your City: Glasgow
- Your Slack Name: Tara

# Homework Details

- Module: HTML/CSS
- Week: Two

# Notes

- What did you find easy?

- What did you find hard?

  I'm still getting to grips with flexbox and am finding there's a lot of trial and error involved before I'm happy with the outcome. 

- Any other notes?
 I'm not quite happy with the way the icons/cards section and the footer below are lined up!


css/style.css
```diff
@@ -17,3 +17,119 @@ body {
...
+li:first-of-type {
+    font-weight:700;
+}
+
+.nav-links {
+    display: flex;
+    justify-content: space-between;
+} 
+
+a:link, a:visited {
```
bonboh, line 53:
```suggestion
.topnav a {
```
It looks like we can apply this style to all `a` elements, so we can leave out `:link` and `:visited`. We can also scope these styles to only the topnav section by changing the selector to `.topnav a`, in case there are other `a` elements on the page

css/style.css
```diff
@@ -17,3 +17,119 @@ body {
...
+    display:flex;
+    justify-content: space-around;
+
+}
+
+.icons {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+}
```
bonboh, line 96:
```suggestion
```
These child elements are already stacked vertically so we don't need to add flex column here. We can also centre the text by adding `text-align: center;` to the `.icon-container` rule instead

index.html
```diff
@@ -9,11 +9,65 @@
...
+            </ul>
+        </nav>
+    </header>
 
-    <!-- Add your HTML markup here -->
-    <!-- Remember: Use semantic HTML tags like <header>, <main>, <nav>, <footer>, <section> etc -->
-    <!-- All the images you need are in the 'img' folder -->
+<main>
+    <section>
+        <div class="hero">
```
bonboh, line 31:
```suggestion
    <section class="hero">
```
It looks like we don't need to have both the section and div, so we can move the class to `section` and remove the `div`

index.html
```diff
@@ -9,11 +9,65 @@
...
+            <button>Learn More</button>
+        </div>
+    </section>
 
+
+    <section class="icon-cards">
+        <h2>Everyone needs a little Karma.</h2>
+        <div class="icon-container">
+            <div class="icons">
+                <img src="img/icon-devices.svg" alt="Various devices illustration" width="100" height="auto">
```
bonboh, line 43:
```suggestion
                <img src="img/icon-devices.svg" alt="Various devices" width="100" height="auto">
```
We don't need to add "illustration" in the `alt` description as we know the context is an image already. Same applies to the other `img` elements below

index.html
```diff
@@ -9,11 +9,65 @@
...
+                <div class="caption"><p>Internet for all devices</p></div>
```
bonboh, line 44:
```suggestion
                <p class="caption">Internet for all devices</p>
```
We don't need to wrap the `p` element in a `div` here, so we can remove the `div` and apply the class on the `p` instead

bonboh (commented):
Great job @Tara-Q! Getting it to look nice on mobile would be a bonus too. I've left some comments and tips for you to read, but you don't need to do anything else 👍

css/style.css
Tara-Q: @bonboh thanks so much for your comments! I'll add some media queries and go back through and implement your suggestions, i definitely knew there would be some extra/redundant stuff in there :/

PR #232, opened 2021-06-18
Scot-Class-5/Karma-Webpage

sPuoja:

# Your Details

- Your Name: Puoja Rajen
- Your City: Glasgow
- Your Slack Name: Puoja

# Homework Details

- Module: HTML/CSS
- Week: 2

# Notes

- What did you find easy?
- html codes

- What did you find hard?
- css and aligning the items

- flexbox

- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -4,6 +4,7 @@
 body {
     font-family: 'Roboto', sans-serif;
     -webkit-font-smoothing: antialiased;
+    font-size: 16px;
 }
 
 /**
@@ -15,5 +16,8 @@ body {
  * 
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
+
+ /* header section */
+
 
 
--- a/css/style.css
+++ b/css/style.css
@@ -18,6 +18,160 @@ body {
  */
 
  /* header section */
+ header{
+     display: flex;
+     justify-content: space-evenly;
+     align-content: center;
+ }
+
+ header img{
+     height: 2.5em;
+     padding: 1rem ;
+   
+}
+
+/* Nav bar */
+
+
+ nav ul{
+     display: flex;
+     list-style: none;
+     justify-content: space-evenly;
+     width: 50rem;
+     padding: 1rem 0rem;
+
+ }
+
+ nav ul a{
+     text-decoration: none;
+     font-size: 1.5rem;
+     color:rgb(197, 189, 189);
+ }
+
+nav ul a:hover{
+    color:rgb(240, 104, 13);
+}
+
+li:first-of-type a{
+    color: rgb(105, 102, 102);
+    font-weight: 500;
+}
+x-special/nautilus-clipboard
+
+/*Main content*/
+.hero{
+  background-position: center;
+  background-repeat: no-repeat;
+  background-size: cover;
+  position: relative;
+
+}
+
+.hero img{
+    display: flex;
+    justify-content: center;
+    padding-left: 1rem;
+    width: 102vh;
+    height: 100%;
+}
+
+
+.hero-text{
+    color: white;
+    position: absolute;
+    top: 30%;
+    left: 47%;
+    transform: translate(-50%, -50%);
+    text-align: center;
+}
+
+.hero-text h1 {
+    font-size: 2.5rem;
+    font-weight: 300;
+}
+
+.hero-text h2{
+    font-size: 1.5rem;
+    font-weight: 300;
+}
+
+.hero button{
+    background-color:rgb(240, 104, 13) ;
+    padding: 1rem;
+    border: none;
+    border-radius: 10px;
+}
+
+/* Icon Section */
+
+.icon img{
+width: 15rem;
+height: 5rem;
+}
+
+.icon-container{
+    display: flex;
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -16,4 +17,161 @@ body {
...
+
+ }
+
+ nav ul a{
+     text-decoration: none;
+     font-size: 1.5rem;
+     color:rgb(197, 189, 189);
+ }
+
+nav ul a:hover{
```
bonboh, line 51:
```suggestion
nav ul a:focus,
nav ul a:hover{
```
When adding hover, we should also add focus so keyboard users can use the `Tab` key to jump to this and also get the same style

css/style.css
```diff
@@ -16,4 +17,161 @@ body {
...
+    color:rgb(240, 104, 13);
+}
+
+li:first-of-type a{
+    color: rgb(105, 102, 102);
+    font-weight: 500;
+}
+x-special/nautilus-clipboard
```
bonboh, line 59:
```suggestion
```
Looks like this was accidentally left in!

css/style.css
```diff
@@ -16,4 +17,161 @@ body {
...
+
+/*Main content*/
+.hero{
+  background-position: center;
```
bonboh, line 63:
```suggestion
.hero{
  height: 40rem;
  background-image: url("../img/first-background.jpg");
  background-position: center;
```
Instead of putting the background image in HTML, we can put it in here and give the element a specific height

css/style.css
```diff
@@ -16,4 +17,161 @@ body {
...
+
+
+.hero-text{
+    color: white;
+    position: absolute;
+    top: 30%;
+    left: 47%;
+    transform: translate(-50%, -50%);
+    text-align: center;
+}
```
bonboh, line 86:
```suggestion
.hero{
    color: white;
    display: flex;
    flex-direction: column;
    justify-content: center;
    align-items: center;
    text-align: center;
}
```
If we delete the `<div class="hero-text">` in the HTML like my other suggestion, we can use flexbox to align the text and button in the centre which is a better way than using `position: absolute`

css/style.css
```diff
@@ -16,4 +17,161 @@ body {
...
+
+.icon img{
+width: 15rem;
+height: 5rem;
+}
+
+.icon-container{
+    display: flex;
+    justify-content: center;
+    flex-direction: row;
```
bonboh, line 115:
```suggestion
```
`flex-direction` is `row` by default, so we can remove it. We usually remove code that has no effect

css/style.css
```diff
@@ -16,4 +17,161 @@ body {
...
+    text-align: center;
+
+}
+.icon-row{
+    padding: 1rem 2rem;
+}
+
+.icon-row p{
+    width: auto;
```
bonboh, line 124:
```suggestion
```
Here, `width` is `auto` by default so we can remove it as it has no effect

css/style.css
```diff
@@ -16,4 +17,161 @@ body {
...
+    color: rgb(46, 45, 45);
+    font-weight: 400;
+    
+}
+
+.icon{
+    width: auto;
+    margin: 0 auto;
+    padding-top: 4rem;
+    box-shadow: 0px 1px 0px rgb(189, 187, 187);
```
bonboh, line 142:
```suggestion
    border-bottom: 1px solid rgb(189, 187, 187);
```
For a single line, we usually use border instead of box-shadow, so we could change it here

css/style.css
```diff
@@ -16,4 +17,161 @@ body {
...
+
+footer{
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+}
+
+footer section{
+    display: flex;
+}
```
bonboh, line 157:
```suggestion
```
The social media icons are already horizontal so we don't need to use flex here

css/style.css
```diff
@@ -16,4 +17,161 @@ body {
...
+
+footer img{
+    margin: 0.5rem;
+    height: 1.5rem;
+    padding: 1rem 1rem;
```
bonboh, line 162:
```suggestion
    padding: 1rem;
```
We usually use the shorthand so we can change it to just `1rem`

index.html
```diff
@@ -10,10 +10,59 @@
...
+                <li><a href="#">Help</a></li>
+                <li><a href="#">Login</a></li>
+            </ul>
+        </nav>
+    </header>
+ 
+    <main>
+        <div class="hero">
+            <img src="img/first-background.jpg">
+            <div class="hero-text">
```
bonboh, line 30:
```suggestion
```
We can remove these, like I mentioned in my CSS suggestions

index.html
```diff
@@ -10,10 +10,59 @@
...
+            <h1>Introducing Karma</h1>
+            <h2>Bring WiFi with you, everywhere you go.<h2>
```
bonboh, line 32:
```suggestion
            <p>Bring WiFi with you, everywhere you go.<p>
```
This doesn't seem like a heading to me — there is already a heading above it — so we could change this to a `p`

bonboh (commented):
Good job @sPuoja! I've left a few comments and tips for you to read, but you don't need to do anything else 👍

PR #233, opened 2021-06-21
Scotland-Class5-Megumi Yanaka-HTML-CSS-Coursework-

MegumiYanaka:
- Your Name: Megumi
- Your City: Scotland
- Your Slack Name: Megumi

# Homework Details

- Module: HTML-CSS-Coursework
- Week: 2

# Notes

- What did you find easy? : Putting just title and sentence and images are fine.

- What did you find hard? : Coding mobile first web site was difficult for me, so I coded only PC site firstly.

- What do you still not understand? : Using flex box and responsive web design.

- Any other notes?  I regretted I coded PC site firstly. I don't know good way how to turn this page into mobile web site.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -16,4 +16,129 @@ body {
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
 
+/* This is Header (Navigation & Hero )*/
+
+/* Navigation */
+nav{
+    display:flex;
+    justify-content: center;
+    align-items: center;
+    height:140px;
+    width: 80%;
+    margin: auto;
+    font-size: 2.5rem;
+    }
+
+a{
+    height: 60px;
+    margin-left: 80px;
+    text-decoration: none;
+    color: rgb(65, 63, 63);
+    }
+
+a:hover{
+    color: rgb(250, 115, 4);
+    }
+
+.logo{
+    width:50px;
+    height:auto;
+    margin-right: auto;
+    margin-left: 0;
+    }
+
+.logo img{
+    width: 100%;
+    height: auto;
+    }
+
+ /* Hero style */
+.hero-image{
+    background-image: url(../img/first-background.jpg);
+    background-position: center;
+    background-repeat: no-repeat;
+    background-size: cover;
+    width: 100%;
+    height: 900px;
+    position:relative;
+}
+
+.hero-text{
+    text-align:center;
+    position: absolute;
+    color: white;
+    top:20%;
+    left:40%;
+}
+
+.hero-text h1{
+    font-size: 5rem;
+}
+
+.hero-text p{
+    font-size: 3rem;
+}
+
+button {
+    background-color: #f2750e; /* Orange */
+    border: none;
+    border-radius: 5px;
+    color: white;
+    font-size: 2rem;
+    padding: 20px 30px;
+    text-align: center;
+    text-decoration: none;
+  }
+
+/*This is main*/
+
+main h2{
+    margin-top: 100px;
+    font-size:4rem;
+    color: grey;
+    text-align: center;
+}
+
+article{
+    display: flex;
+    justify-content: space-around;  
+}
+
+article div{
+    text-align: center;
+    font-size: 3rem;
+    margin-top: 5rem;
+    margin-bottom: 10rem;
+}
+    
+article img{
+    width:60%;
+}
+
+.devices{
+    width: 100%;
+}
+
+/* This is footer*/
+
+footer {
+    font-size: 2.5rem;
+    text-align: center;
+    margin-top: 5rem;
+}
+
+.sns-icon{
+    display: flex;
+    justify-content: center;
+    height:40px;
+}
+
+.sns-icon img{
+    height:100%;
+    padding:2rem;
+}
+footer p{
+    margin-top: 6rem;
+    color: grey;
+}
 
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -16,4 +16,129 @@ body {
...
+
+/* Navigation */
+nav{
+    display:flex;
+    justify-content: center;
+    align-items: center;
+    height:140px;
+    width: 80%;
+    margin: auto;
+    font-size: 2.5rem;
```
bonboh, line 29:
```suggestion
    font-size: 1rem;
```
All the text on this page is quite big! We could reduce the values of all font-sizes in this file

css/style.css
```diff
@@ -16,4 +16,129 @@ body {
...
+    }
+
+a{
+    height: 60px;
+    margin-left: 80px;
+    text-decoration: none;
+    color: rgb(65, 63, 63);
+    }
+
+a:hover{
```
bonboh, line 39:
```suggestion
a:focus,
a:hover{
```
When we add hover, we should also add focus so keyboard users who use the `Tab` key to jump to this link will get these styles too

css/style.css
```diff
@@ -16,4 +16,129 @@ body {
...
+    color: rgb(250, 115, 4);
+    }
+
+.logo{
+    width:50px;
+    height:auto;
```
bonboh, line 45:
```suggestion
```
`height` is `auto` by default here so this doesn't do anything. We usually remove code that has no effect

css/style.css
```diff
@@ -16,4 +16,129 @@ body {
...
+    position:relative;
+}
+
+.hero-text{
+    text-align:center;
+    position: absolute;
+    color: white;
+    top:20%;
+    left:40%;
+}
```
bonboh, line 72:
```suggestion
.hero-text{
    background-image: url(../img/first-background.jpg);
    background-position: center;
    background-repeat: no-repeat;
    background-size: cover;
    width: 100%;
    height: 900px;
    
    text-align:center;
    color: white;
    
    /* use flexbox to centre the text and buttons */
    display: flex;
    flex-direction:column;
    align-items: center;
    justify-content: center;
}
```
Instead of using `position: absolute` to position it in the centre, a better way for aligning things in the centre is to use flexbox

css/style.css
```diff
@@ -16,4 +16,129 @@ body {
...
+    font-size: 3rem;
+}
+
+button {
+    background-color: #f2750e; /* Orange */
+    border: none;
+    border-radius: 5px;
+    color: white;
+    font-size: 2rem;
+    padding: 20px 30px;
```
bonboh, line 88:
```suggestion
    padding: 1.25rem 2rem;
```
We should use relative units like rem instead of absolute units like px where possible. Same goes for the rest of this file

css/style.css
```diff
@@ -16,4 +16,129 @@ body {
...
+    text-align: center;
+    text-decoration: none;
```
bonboh, line 90:
```suggestion
```
The text is already centred in a button so we don't need `text-align: center`. `text-decoration: none` only applies to links so we don't need it here for the button

css/style.css
```diff
@@ -16,4 +16,129 @@ body {
...
+
+footer {
+    font-size: 2.5rem;
+    text-align: center;
+    margin-top: 5rem;
+}
+
+.sns-icon{
+    display: flex;
+    justify-content: center;
```
bonboh, line 132:
```suggestion
```
The social media icons are already centred due to the footer's `text-align: center` so we don't need  flexbox here

index.html
```diff
@@ -9,11 +9,66 @@
...
+    <nav>    
+        <a href="#" class="logo"> 
+           <img src="img/karma-logo.svg" alt="logo">
+        </a>     
+        <a href="">Meet Karma</a>
+        <a href="">How it works</a>
+        <a href="">Store</a>
+        <a href="">Blog</a>
+        <a href="">Help</a>
+        <a href="">Log in</a>
```
bonboh, line 30:
```suggestion
        <ul>
            <li><a href="">Meet Karma</a></li>
            <li><a href="">How it works</a></li>
            <li><a href="">Store</a></li>
            <li><a href="">Blog</a></li>
            <li><a href="">Help</a></li>
            <li><a href="">Log in</a></li>    
        </ul>
```
In a `nav`, there's usually a list of links, and because it's a list we usually put them in an HTML list. We can then use CSS to remove the bullet points and change the list to inline boxes

index.html
```diff
@@ -9,11 +9,66 @@
...
+    </nav>
+
+    <!--Hero-->
```
bonboh, line 33: This hero section seems like it's part of the main section, not the header, so we should move it into the `main` element

index.html
```diff
@@ -9,11 +9,66 @@
...
+    <section class="hero-image"></section>
+        <div class="hero-text">
```
bonboh, line 35:
```suggestion
    <section class="hero-text">
```
With the flexbox suggestions I gave for the hero section, we don't need both a section and div so we can delete the div

index.html
```diff
@@ -9,11 +9,66 @@
...
+        </div>
+    </section>
+</header>
+
+    <!--Main-->
+<main>
+    <h2>Everyone needs a little Karma.</h2>
+    <article>
+     <div>
+        <img src="img/icon-devices.svg">
```
bonboh, line 48:
```suggestion
        <img src="img/icon-devices.svg" alt="Various computer devices">
```
`img` elements should have an `alt` description for screen reader users. Same applies to the rest of the images

bonboh (commented):
Good effort @MegumiYanaka! I've left some suggestions and tips for you to read, but you don't need to do anything else 👍

PR #238, opened 2021-06-23
Scotland class 5 - Houda - Karma clone

bonboh:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -15,5 +15,147 @@ body {
  * 
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
+ 
+ nav  {
+    display:flex;
+    justify-content: flex-end;
+     padding-top: 60px;
+     padding-left: 20px;
+     
+ } 
 
+.nav{
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 50%;
+    
+}
+#logo{
+   margin-right:39em;
+    
+    
+    
+}
+.nav-list {
 
+    display: flex;
+    margin-right: 5rem;
+  }
+  .nav-list a {
+    display:flex;
+    justify-content: flex-end;
+    display: block;
+    font-size: 1rem;
+    padding: 1rem;
+    text-decoration: none;
+    color: #2C3539;
+    
+  }
+  ul {
+    list-style-type: none;
+    margin: 0;
+    padding: 0;
+  }
+  section{
+    background-image: url(/home/coder/GitHub/HTML-CSS-Coursework-Week2/img/first-background.jpg);
+      display:flex;
+      flex-direction:column;
+      justify-content: center;
+      align-items: center;
+      padding-top: 100px;
+      padding-bottom: 100px;
+      
+      margin-left: 30px;
+      margin-right: 10px;
+       
+      color:white;
+    }
+    h1{
+        font: 40px Arial, sans-serif;
+        color:antiquewhite;
+
+    }
+  .icons{
+      font-size: 20px;
+      color:#2C3539;
+      justify-content: center;
+      text-align:center ;
+      padding-top: 50px;
+
+  }
+  .icons-img{
+      display:flex;
+      flex-direction:row;
+      justify-content: center;
+      align-items: center;
+      margin-left: 10px;
+      padding-top: 30px;
+
+  }
+  .device{ display:flex;
+    flex-direction:column;
+    justify-content: center;
+    align-items: center;
+
+    padding-left: 30px;
+    
+    width: 20%;
+      
+  }
+  .coffee{
+    display:flex;
+    flex-direction:column;
+    justify-content: center;
+    align-items: center;
+
+    padding-left: 30px;
+    
+    width: 20%;
+
+  }
+  .petrol{ display:flex;
+    flex-direction:column;
+    justify-content: center;
+    align-items: center;
+
+    padding-left: 30px;
+    background-color:antiquewhite
+    width: 20%;
+
+  }
+  button{
+      
+      width: 10%;
+      height: 5vh;
+      font-size: 11px;
+      border-radius: 5px;
+      background-color:chocolate;
+      
+    
+  }
+  button a{
+    color:antiquewhite;
+    text-decoration: none;
+  }
+
+   
+  .media{
+       
+      margin-left: 50%;
+  }
+  .media-p{
+    padding-top: 60px;
+      margin-left: 48%;
+  }
+  footer{
+      
+      
+      justify-items: center;
+      text-align: center;
+      margin-left:3% ;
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -15,5 +15,147 @@ body {
  * 
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
+ 
+ nav  {
+    display:flex;
+    justify-content: flex-end;
```
bonboh, line 21:
```suggestion
    justify-content: space-around;
```
We can change this to `space-around`, and delete the `#logo`'s `margin-right:39em;` below on line 34, for a better way to put space between the logo and the menu links

css/style.css
```diff
@@ -15,5 +15,147 @@ body {
...
+     padding-top: 60px;
```
bonboh, line 22:
```suggestion
     padding-top: 3.75rem;
```
We should use relative units like rem instead of px where possible. This applies to the other px's in the rest of this file

css/style.css
```diff
@@ -15,5 +15,147 @@ body {
...
+     padding-left: 20px;
+     
+ } 
 
+.nav{
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 50%;
```
bonboh, line 31:
```suggestion
    width: 100%;;
```
`top` and `left` only have an effect when we've used the `position` CSS property, but since we haven't used it then we can remove `top` and `left` as they're not doing anything.

`height: 50%` also doesn't have an effect so we should remove it. We usually remove any code that doesn't have any effect

css/style.css
```diff
@@ -15,5 +15,147 @@ body {
...
+    
+}
+.nav-list {
+
+    display: flex;
+    margin-right: 5rem;
+  }
+  .nav-list a {
+    display:flex;
+    justify-content: flex-end;
```
bonboh, line 47:
```suggestion
```
`display:flex` only works when the element has child elements. the `a` element doesn't have any child elements, so `display:flex` doesn't do anything and we can remove it

css/style.css
```diff
@@ -15,5 +15,147 @@ body {
...
+    color: #2C3539;
+    
+  }
+  ul {
+    list-style-type: none;
+    margin: 0;
+    padding: 0;
+  }
+  section{
+    background-image: url(/home/coder/GitHub/HTML-CSS-Coursework-Week2/img/first-background.jpg);
```
bonboh, line 61:
```suggestion
    background-image: url(../img/first-background.jpg);
```
We should use paths that are relative, so something like this. A relative path means I don't need to have the same computer as you `/home/coder/GitHub/HTML-CSS-Coursework-Week2/img/first-background.jpg`

css/style.css
```diff
@@ -15,5 +15,147 @@ body {
...
+       
+      color:white;
+    }
+    h1{
+        font: 40px Arial, sans-serif;
+        color:antiquewhite;
+
+    }
+  .icons{
+      font-size: 20px;
```
bonboh, line 80:
```suggestion
      font-size: 1.25rem;
```
It's important that font sizes use relative units, so we can update this to use rem

css/style.css
```diff
@@ -15,5 +15,147 @@ body {
...
+      color:#2C3539;
+      justify-content: center;
```
bonboh, line 82:
```suggestion
```
`justify-content` only works when the element also has `display: flex` on it. There's no `display: flex` on this element so we can remove `justify-content` as it's not doing anything

css/style.css
```diff
@@ -15,5 +15,147 @@ body {
...
+
+  }
+  .device{ display:flex;
+    flex-direction:column;
+    justify-content: center;
+    align-items: center;
+
+    padding-left: 30px;
+    
+    width: 20%;
```
bonboh, line 103:
```suggestion
  .device{
    padding-left: 30px;
    width: 20%;
```
The `img` icon and the `p` text children are already displayed vertically, so we don't need to do flex column here. The same goes for the `.coffee` and `.petrol` styles

We can also make the text centred instead by adding `text-align: center;` to the `.icons-img` rule above

css/style.css
```diff
@@ -15,5 +15,147 @@ body {
...
+
+    padding-left: 30px;
+    background-color:antiquewhite
+    width: 20%;
+
+  }
+  button{
+      
+      width: 10%;
+      height: 5vh;
```
bonboh, line 130:
```suggestion
      padding: 1rem;
```
Instead of adding a specific width and height, we can use padding instead to add space around the button text, which is more commonly done

index.html
```diff
@@ -9,11 +9,87 @@
     <link rel="stylesheet" href="css/style.css">
     <link rel="shortcut icon" type="image/x-icon"  href="favicon.ico">
 </head>
-<body>
+
+<header>
```
bonboh, line 13: This `header` element should go in the `body`, not `head`. All HTML code goes in `body`. The `head` is for other things not HTML

index.html
```diff
@@ -9,11 +9,87 @@
...
+    
+    
+    <nav id="navbar" class="nav">
+        <img id="logo" src="img/karma-logo.svg" alt=# width="30" height="50" />
```
bonboh, line 17:
```suggestion
        <img id="logo" src="img/karma-logo.svg" alt="Karma logo" width="30" height="50" />  
```
All images should have an `alt` description which describes the image. Same applies to the other images in this file

index.html
```diff
@@ -9,11 +9,87 @@
...
+
+</header>
+  <body>
+  <section>
+     <h1>Introduction Karama </h1>
+     <p>  bring wifi with you,every where you go </p>
+     
+     <button >
+     <a href= "#"> learn more  </a>
+     </button>
```
bonboh, line 48:
```suggestion
     <button >
       learn more
     </button>
```
It doesn't really make sense to have a link inside a button — it should be either a link or a button

index.html
```diff
@@ -9,11 +9,87 @@
...
+
+</div> 
+  
 
     <!-- Add your HTML markup here -->
     <!-- Remember: Use semantic HTML tags like <header>, <main>, <nav>, <footer>, <section> etc -->
     <!-- All the images you need are in the 'img' folder -->
 
 </body>
+<div>
```
bonboh, line 83:
```suggestion
<footer>
```
It looks like this is a footer, so we can change the element to `footer`. It should also be inside the `body` because all HTML code goes inside the `body`

bonboh (commented):
Good effort @kenda121! I've left a few comments and tips for you to read, but there's nothing else you need to do 👍

PR #288, opened 2021-11-16
London - class - 8 - Html/ CSS - Mohammad Alamin - Week 2 and Week 3

Alamin-eng:
Desktop and Mobile Extended version of the Karma page and its Store page

# Your Details

- Your Name: Mohammad Alamin
- Your City: London
- Your Slack Name: Mohammad Alamin - trainee

# Homework Details

- Module: Html/CSS
- Week: 2 and 3

# Notes

- What did you find easy? Html/CSS code for desktop version is kind of easy .

- What did you find hard? Styling for the mobile version is a bit difficult, Functional  hamburger bar (not the one that I already have in my page)  for the mobile version would be important  and kind of difficult to code. On Week 3 - custom making radio buttons and check boxes were difficult.

- What do you still not understand? I don't really understand the @media screen sizes and which are best max or min px size!

- Any other notes? I am happy with my final commit, everything looks fine to me


Alamin-eng:
@haack

Alamin-eng:
https://alamin-eng.github.io/HTML-CSS-Coursework-Week2/

index.html
```diff
@@ -15,5 +15,155 @@
     <!-- Remember: Use semantic HTML tags like <header>, <main>, <nav>, <footer>, <section> etc -->
     <!-- All the images you need are in the 'img' folder -->
 
+<!-- Header logo and navbar -->
+    <header class='header'>
```
haack, line 19: Nice one using semantic tags 🏅

index.html
```diff
@@ -15,5 +15,155 @@
...
+        <!-- svg logo of Hamburger menu -->
+        <a class='hamburgerMenu' href="#" >
+            <svg width="24" height="16" viewBox="0 0 24 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+                <path fill-rule="evenodd" clip-rule="evenodd" d="M0 1C0 0.447715 0.447715 0 1 0H23C23.5523 0 24 0.447715 24 1V1.28572C24 1.838 23.5523 2.28572 23 2.28572H1C0.447716 2.28572 0 1.838 0 1.28572V1ZM0 7.85714C0 7.30485 0.447715 6.85713 1 6.85713H23C23.5523 6.85713 24 7.30485 24 7.85713V8.14285C24 8.69513 23.5523 9.14285 23 9.14285H1C0.447716 9.14285 0 8.69513 0 8.14285V7.85714ZM1 13.7143C0.447715 13.7143 0 14.162 0 14.7143V15C0 15.5523 0.447716 16 1 16H23C23.5523 16 24 15.5523 24 15V14.7143C24 14.162 23.5523 13.7143 23 13.7143H1Z" fill="black"/>
+            </svg>
+        </a>
+        <!-- End of Hamburger menu  -->
+
+        <!-- Navbar -->
+          <div>
```
haack, line 40: You can use the `<nav>` semantic element here

index.html
```diff
@@ -15,5 +15,155 @@
...
+                  fill="url(#b)" />
+              </svg>
+            
+        </div> <!-- end of social media's svg div -->
+
+      <!-- Copywrite information and My purpose of editing this page   -->
+        <p style="color:rgb(160, 157, 157)"> &copy; Karma Mobility Inc.</p>
+        <p style="color:rgb(160, 157, 157)"> Edited with love by Mohammad Alamin as part of week 2 assignment</p>
```
haack, line 166: Rather than putting `color:rgb(160, 157, 157)` in the tag here, you can put it in a class and reuse it in the line above too

index.html
```diff
@@ -15,5 +15,155 @@
...
+                    <h2>anymore!</h2>
+                  </em> 
+                <button>Get Karma today</button>
+            </div>
+        </div>
+    <!-- End of extended-article  -->
+
+    </div><!-- End of content div -->
+
+<!-- hr and line break used -->
```
haack, line 115: I'd suggest comments like these shouldn't be necessary. As much as possible try and make your code as clear as possible so that you don't need to use comments as much.

css/style.css
```diff
@@ -16,4 +17,285 @@ body {
...
+    color:white;
+    font-weight: 100;
+ }
+
+/* Main content's texts  */
+ .main>h1{
+     font-size: 3.5em;
+     letter-spacing:3px;
+ }
+ .h1Paragraph{
```
haack, line 93: Try to keep the naming style of classes consistent. Usually for css classes people used what's called `kebab-case` (which you have done below).

haack (commented):
Great job! Good use of flex box and semantic tags 👍 

Don't worry if the media queries don't yet make sense because we haven't learnt them yet, but well done for having a go

Alamin-eng:
Hi Alex. I have committed and pushed both pages for Week 3 assignment  @haack

haack (commented):
Nice! Well done on styling the radio buttons and checkbox too.

I've noticed your class names have improved since last week, well done! Remember to keep on being "perfectionist" and consistent when you are writing code. It might not seem helpful at the moment, but it is really important! CYF has a style-guide if you want to read more about this: https://syllabus.codeyourfuture.io/guides/code-style-guide

I've noticed you've manually set the height and width for lots of elements. This can be a lot of effort and difficult to tweak for each one. Instead of setting a fixed height or width, try just setting internal sizing for the element (for example `padding`) and let that determine the height/width of your component automatically.

Lastly try to reuse css classes more! This will mean you need much less css code (which will make your life much easier)! For example, perhaps try creating a class the set the style for your text fields and use that for each one (rather than setting custom styles for each individual field). In general, try to look for where there are common styles in your page and write classes that you can use in multiple places

PR #290, opened 2021-11-17
london8-week2

Alirezabg:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


Store.html
```diff
@@ -0,0 +1,154 @@
...
+        </ul>
+    </div>
+
+  </header>
+  <main>
+      <section class="store-section">
+
+         <DIV>
+            <img class="store-img" src="img/store-image_by-andrew-neel-unsplash.jpg" alt=""/>
+
```
ellietms, line 56: for images, always wrap them inside a div (as their parent) then give that div a width of 100%, and give the img (which was wrapped with div) the width you want(usually by percentage).
you should have a class for the div and just have a width for the class of this img ( you should not use height)

Store.html
```diff
@@ -0,0 +1,154 @@
...
+        </DIV>
+        <DIV class="store-form">
```
ellietms, line 58: you wrapped everything inside of this div( div which has store-form class) and you have different classes for most of the elements inside of this div, inside these classes: `form-header column-form row-form` you used `display: flex,flex-direction, etc`.
instead of doing this and the right way for using the flex-box is you should always write `display: flex` inside of a class of the parent which means you just need to use display: flex inside of the wrapper class (div) and delete the display: flex for all of the element's classes which were wrapped by this div and just specified the flex-direction or other flex features you want in the classes of the elements.

Store.html
```diff
@@ -0,0 +1,154 @@
+<!DOCTYPE html>
```
ellietms, line 1: @Alirezabg  

🥳 overall well done for using :
1) semantic form elements
2) using flex-boxes 

🤔 What you need to improve:
1) well-named CSS classes
2) using HTML nesting and indentation
3) using flex-box for parents elements  and work more on how to use different classes for parent and the child elements

Store.html
```diff
@@ -0,0 +1,154 @@
...
+                    <legend></legend>
+                    <div class="column-form spc">
+                        <label for="firstname">First name *</label>
+                    <input type="text" id="firstname" name="firstname" required>
+                    
+
+                    </div>
+                    <div class="column-form">
+                        <label for="lastname">Last name *</label>
+                        <input type="text" id="lastname" name="lastname" required>
```
ellietms, line 71: you can add some classes for your input, to make it look a little bit better( maybe make it a little bit smaller and change the width of it?)

PR #294, opened 2021-11-17
London class 8 - Craig D'silva - HTML/CSS - week 2

craig-dsilva:

# My Details
Name: Craig D'Silva
City: London
Slack Name: Craig - Trainee

# Homework Details
Module: HTML/CSS
Week: Week 2

# Notes
What did you find easy?
The positioning part was easy as I used flexbox which adjusted it's size according to the viewport. 

What did you find hard?
The hardest part was to make the image fit in the viewport as it gets covered when I make the screen small. The next thing that was hard was adding the image and quote together side by side. But I solved this issue by using a flexbox.

What do you still not understand?
I still can't figure out how the image should remain within the viewport when the user decreases the width of the browser window.

Any other notes?
I have tried to make this website as responsive as possible. I tried to use relative units (percentages) wherever possible and I'm not sure if I have done it right. Please give feedback for this project.


css/store.css
```diff
@@ -0,0 +1,119 @@
+:root {
+  --form-colour: #828282;
+}
```
LucyMac, line 3: Nice variable :)

css/store.css
```diff
@@ -0,0 +1,119 @@
...
+
+.form-colour-text {
+  /* Order colour selector text */
+  margin-bottom: 2%;
+}
+
+.form-colour {
+  /* Colour section of the form */
+  margin-bottom: 10%;
```
LucyMac, line 90: In general, I'd avoid using % for padding and margin because these often need to be consistent / predictable. Use `rem` instead for a fixed size that is still relative to the dimensions of the page.

css/store.css
```diff
@@ -0,0 +1,119 @@
...
+  color: #f15b2a;
+}
+
+.image {
+  background-image: url("../img/store-image_by-andrew-neel-unsplash.jpg");
+  background-repeat: no-repeat;
+  background-size: cover;
+  padding: 0 25% 60%;
+  margin-bottom: -8px;
+}
```
LucyMac, line 119: Your store page is wider than the screen allows, which is why it scrolls horizontally. This is because you have not set the width for each half. The form should be 50% and so should the image. 

Check your `.form` class on line 17 - your margin may also be an issue as it will add to the 50%!

css/style.css
```diff
@@ -16,4 +15,210 @@ body {
...
+  margin: 15% 12% 25px;
+}
+
+.quote-orange {
+  color: #f15b2a;
+  font-weight: 600;
+}
+
+.quote-button {
+  margin: 0 40% 10%;
```
LucyMac, line 146: This margin is forcing the button to be too small and wrap onto several lines. 
You should need any left and right margin - just size the button using padding, and center it using flexbox.

css/style.css
```diff
@@ -16,4 +15,210 @@ body {
...
+.social-icon {
+  border: 1px solid #e1e3e6;
+  border-radius: 50px;
+  padding: 15px;
+  margin-right: 25px;
+}
+
+.social-icon-image {
+  width: 25px;
+  height: 25px;
```
LucyMac, line 180: Add `display: block` here to make the circles round.

index.html
```diff
@@ -1,19 +1,140 @@
...
+      </section>
+
+      <!-- Quote subsection -->
+      <section class="quote">
+        <!-- Used flexbox so that image and quote can go side by side -->
+        <div class="quote-image-box"></div>
+
+        <div class="quote-text-box">
+          <p class="quote-text">
+            <i>
```
LucyMac, line 93: You could use the `blockquote` tag here instead, and just make it italic in your CSS,

LucyMac (commented):
Nice work @craig-dsilva, a few comments for you to look into.
Watch out with using % units for spacing, and also be aware that horizontal margins add space on the outside of elements, which messes with the layout of the page. Try and favour padding instead when you can.

css/store.css
craig-dsilva: Thanks

css/store.css
craig-dsilva: I was so confused about CSS units. Thanks for the advice.

craig-dsilva:
Thanks @LucyMac  for the feedback. It's been really helpful.

PR #295, opened 2021-11-17
London class 8 - Harsheek Thanki - HTML/CSS - Week 1

HThanki:
Hi @jackhclee, please find my Karma site PR for review

# Your Details

- Your Name: Harsheek Thanki
- Your City: London
- Your Slack Name: Harsheek Thanki

# Homework Details

- Module: HTML/CSS
- Week: 1

# Notes

- What did you find easy? Structuring the main HTML parts

- What did you find hard? Breaking the smaller HTML parts into divs and articles

- What do you still not understand? The main difference between padding and margin. and when to use which I think padding adds empty space within the element and margin is outside the element.

- Any other notes?
No. Time for sleep :)


css/style.css
```diff
@@ -16,4 +15,110 @@ body {
...
+}
+
+article {
+  padding: 100px;
+}
+
+.hero {
+  background: url(../img/first-background.jpg) no-repeat;
+  background-size: cover;
+  height: 250px;
```
LucyMac, line 71: You might find that using internal padding on the hero is a more flexible way of controlling its size, than setting a fixed height. It won't make too much of a difference now, but when we start adding responsiveness to the page, it might make things easier :)

css/style.css
```diff
@@ -16,4 +15,110 @@ body {
...
+h1 {
+  font-weight: normal;
+  font-size: 50px;
+}
+
+p {
+  font-size: 20px;
+}
+
+.second-section {
```
LucyMac, line 93: Beware of naming things by their position on the page. The order of elements could change with time, or on different screen sizes. Think instead "what does this section do?" and name it that way.

css/style.css
```diff
@@ -16,4 +15,110 @@ body {
...
 
+footer p {
+  font-size: 15px;
+  padding-bottom: 10px;
+}
+
+footer img {
+  width: 15px;
+  border: 1px solid #bdc3c7;
+  border-radius: 100px;
```
LucyMac, line 118: Your icons are being rounded by the border. Don't apply the border directly on the image - instead, wrap it inside a `div` and apply the border and border-radius to that.

index.html
```diff
@@ -1,19 +1,82 @@
...
+        <div class="second-section-columns">
+          <div class="col-1">
+            <img src="img/icon-devices.svg" />
+            <p>Internet for all devices</p>
+          </div>
+          <div class="col-2">
+            <img src="img/icon-coffee.svg" />
+            <p>Boost your productivity</p>
+          </div>
+          <div class="col-3">
```
LucyMac, line 51: You're not using these classes anywhere in your CSS, so you don't need them here. 
Additionally, each of these 3 blocks is basically exactly the same - so why use a different class? Classes are made to be re-usable. Always re-use code as much as possible!

LucyMac (commented):
Very nice Harsheek, your code it's super neat and well organised, I like it! 🏅 
A couple of tweaks to improve a little but on the whole great work.

css/style.css
HThanki: Thanks @LucyMac, appreciate the feedback. "second" section renamed to "information" section to reflect the section is the information about the Karma product

index.html
HThanki: Thanks Lucy, unused classes have been removed

css/style.css
HThanki: Thanks, border and border-radius added to div rather than on image as suggested

css/style.css
```diff
@@ -103,20 +102,56 @@ p {
...
+
+.quote-rectangle {
+  background-color: E0633A;
+  border: 5px solid blueviolet;
+}
+
+.quote-text {
+  padding: 20px 40px 0px;
+  font-style: italic;
+}
```
jackhclee, line 125: Great, I see CSS classes are better named and I guess you are writing some CSS class for store.html as well. Just remember to split them to separate CSS files to improve CSS coverage (i.e. improve better code utilisation)

PR #299, opened 2021-11-18
London Class 8 - Delroy Gayle - HTML/CSS - Week 2

DelroyGayle:
This is my coursework regarding the Karma website


css/style.css
```diff
@@ -17,3 +17,325 @@ body {
...
+  .centred {
+    position: absolute;
+    top: 36%;
+    left: 46%;
+    transform: translate(-50%, -50%);
+  }
+
+  .para1 {
+    font-family: Roboto; 
+    font-size:72px;
```
Sam-Martin, line 152: Avoid absolute sizing wherever you can! Say no to `px` say yes to `rem`!

css/style.css
```diff
@@ -17,3 +17,325 @@ body {
...
+  /* Based on https://www.w3schools.com/howto/howto_css_image_text.asp */
+
+  .centred {
+    position: absolute;
+    top: 36%;
+    left: 46%;
+    transform: translate(-50%, -50%);
+  }
+
+  .para1 {
```
Sam-Martin, line 150: What is this class name telling us? That it's the first paragraph? That sounds more like an id than a class name, which we should avoid if possible. When naming and separating your classes think about how to make them as reusable as possible. If you have to specify something only really occurs in one specific scenario, see if you can specify that with a selector that uses semantic elements or parent elements (e.g. `.mainImage > h2` or something.)

css/style.css
```diff
@@ -17,3 +17,325 @@ body {
...
+    font-family: Roboto; 
+    font-size:72px;
+  
+  }
+
+  .para2 {
+    font-family: Roboto; 
+    font-size:36px;
+    position: absolute;
```
Sam-Martin, line 159: Avoid absolute positioning like the plague! It will really get you into trouble. Try and do everything with margins and flexboxes.

css/style.css
```diff
@@ -17,3 +17,325 @@ body {
...
+  border-top-color:gray;
+  background-color: white;
+  text-align: center;
+ }
+  
+ 
+    /* Based on https://www.w3.org/Style/Examples/007/center.en.html */
+
+  .placeInTheCentre {
+    position: absolute;
```
Sam-Martin, line 253: You know what I'm going to say here 🙂

css/style.css
```diff
@@ -17,3 +17,325 @@ body {
...
+  font-size:52px;
+  color:black;
+  padding-top:10rem;
+  padding-left:27rem;
+ }
+
+
+   /* FLEXBOX for three icons*/
+
+   .iconFlexBox {
```
Sam-Martin, line 209: Yay flexboxes! 🎉 nice 👍

css/style.css
```diff
@@ -17,3 +17,325 @@ body {
...
+      object-fit: cover;  
+
+      /* For this code to work well the starting color needs to be black. 
+      If your icon set isn't black you can prepend "brightness(0) saturate(100%)" to your filter property 
+      which will first turn the icon set to black. */
+
+      filter: brightness(0) saturate(100%);
+
+      /* Now convert to Instagram Blue, that is, the target colour is #405DE6; */
+      filter: invert(40%) sepia(78%) saturate(481%) hue-rotate(181deg) brightness(83%) contrast(85%);
```
Sam-Martin, line 326: Clever! But as you say a bit overkill. Better to tweak the SVG itself to change the colour, though that is more complex.

index.html
```diff
@@ -8,12 +8,121 @@
...
+                          <div class="grayBorder">
+                            <img src="img\instagram-icon.svg" alt="Instagram link"; height=20px; width=20px; class="blueInstagram">   
+                          </div>                     
+                        </a>
+                     
+              </div>
+          </div>
+    
+         <div class="copyright">&copy; Karma Mobility, Inc.</div>
+    </div>
```
Sam-Martin, line 113: This indentation doesn't look right. Each level of tag should result in an additional indent like so: 
```
<html>
  <body>
    <footer>
    </footer>
  </body>
</html>
```

Sam-Martin (commented):
The icons don't look right here, they're falling off the screen.
image
You've used a flex box to good effect, but getting rid of margins and using absolute positioning is making your life harder than it needs to be.
Indentation  also needs some work in both the html and the css.
Try getting rid of absolute positioning and working on your indentation and you should be in a good place 👍

Sam-Martin (commented):
The icons don't look right here, they're falling off the screen.
image
You've used a flex box to good effect, but getting rid of margins and using absolute positioning is making your life harder than it needs to be.
Indentation  also needs some work in both the html and the css.
Try getting rid of absolute positioning and working on your indentation and you should be in a good place 👍

index.html
```diff
@@ -8,12 +8,121 @@
...
+                    <li class="navigation-item">
+                      <a class="navigation-link" href="#">Login</a>
+                    </li>
+                    </ul>
+              </nav>
+    </div>
+
+    <div class="mainImage">
+            <div class="centred">
+                <p class="para1">Introducing Karma</p>
```
Sam-Martin, line 53: Are there better tags than `p` that represent the role this text is playing in the page? maybe one of the `h` tags would be a better fit?

DelroyGayle:
Dear @Sam-Martin

Thank you for your comments and recommendations

However I am stumped!
The icons definitely do *not* look like that on my laptop!
See attached.
What do you suggest?

Kind regards
Delroy

Sam-Martin:
Hi @DelroyGayle definitely start by replacing the absolute position styling and see how you go from there.

DelroyGayle:
Thanks for the feedback!

index.html
```diff
@@ -8,12 +8,102 @@
...
     <link rel="stylesheet" href="css/style.css">
     <link rel="shortcut icon" type="image/x-icon"  href="favicon.ico">
+    <title>Introducing Karma</title>
 </head>
 <body>
 
-    <!-- Add your HTML markup here -->
-    <!-- Remember: Use semantic HTML tags like <header>, <main>, <nav>, <footer>, <section> etc -->
-    <!-- All the images you need are in the 'img' folder -->
+    <section class="LogoNavbox">
```
amellie, line 15: I would actually use `<header>` instead of `<section>` since this what this part is all about

index.html
```diff
@@ -8,12 +8,102 @@
...
+                  <img
+                      class="karma-logo"
+                      src="img/karma-logo.svg"
+                      alt="The Logo for Karma"
+                  />
+              </a>
 
+              <!-- see https://www.w3schools.com/css/css_navbar.asp -->
+              
+              <nav>
```
amellie, line 27: Perfect navigation and the re-use of those classes!

index.html
```diff
@@ -8,12 +8,102 @@
...
+                    <a class="navigation-link" href="#">How it Works</a>
+                  </li>
+                  <li class="navigation-item">
+                    <a class="navigation-link" href="#">Store</a>
+                  </li>
+                  <li class="navigation-item">
+                    <a class="navigation-link" href="#">Blog</a>
+                  </li>
+                  <li class="navigation-item">
+                      <a class="navigation-link" href="#">Help</a>
```
amellie, line 42: the indentation is quite off from here onwards

index.html
```diff
@@ -8,12 +8,102 @@
...
+                    </li>
+                    <li class="navigation-item">
+                      <a class="navigation-link" href="#">Login</a>
+                    </li>
+                    </ul>
+              </nav>
+    </section>
+
+    <main class="MainFlexBox">
```
amellie, line 51: Correct!

css/style.css
```diff
@@ -17,3 +17,232 @@ body {
...
+.everyone {
+  font-family: "Roboto", Arial, sans-serif; 
+  font-size: 3.25rem; /* 52px; */
+  color:black;
+  padding-top:10rem;
+ }
+
+.drawaline {
+  width: 90%;
+  /* --grey-light: #e4e4e4; */
```
amellie, line 163: remove unused codes please :)

amellie (commented):
Thank you for taking the time to update your work based on our feedback! It looks great now and just a few comments from me

PR #300, opened 2021-11-18
Karma rough draft

AmarachiOkereke:
Still working on this

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,9 +1,8 @@
-
- /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
+/* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
 
 body {
-    font-family: 'Roboto', sans-serif;
-    -webkit-font-smoothing: antialiased;
+  font-family: "Roboto", sans-serif;
+  -webkit-font-smoothing: antialiased;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -1,19 +1,121 @@
...
+        </div>
+      </div>
+    </header>
+
+    <section>
+      <p class="title">Everyone needs a little Karma.</p>
+      <div class="how-container">
+        <div class="how_it_works">
+          <img
+            src="./img/icon-devices.svg"
```
JoeWeate, line 56: Instead of using padding to separate some centered items, on the "how container" class you can change` justify-content: center` to `space-around`. This will let you delete the left and right paddings.

index.html
```diff
@@ -1,19 +1,121 @@
...
+          />
+          <p>Pay as You Go</p>
+        </div>
+      </div>
+    </section>
+
+    <footer>
+      <p class="help">Join us on</p>
+      <div class="socials">
+        <ul class="help-links">
```
JoeWeate, line 86: Your help-links are inheriting a default left-padding from the `ul` list items. This is why your social media icons aren't completely centered.  You can override that with either `ul { padding: 0;}` or adding `padding:0` to `help-links `

css/style.css
```diff
@@ -16,4 +15,118 @@ body {
...
+
+/* section styling starts here */
+.how-container {
+  display: flex;
+  justify-content: center;
+  flex-direction: row;
+  align-items: center;
+}
 
+.how_it_works {
```
JoeWeate, line 66: Some of your class names use underscores while the others use dashes. I'd stick to dashes as that's more natural to CSS.

css/style.css
```diff
@@ -16,4 +15,118 @@ body {
...
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding: 6rem;
+  margin: 0.5rem;
+}
+
+.title {
```
JoeWeate, line 74: You're giving every `<p>` tag this class except for the actual Title `"Introducing Karma"`. Make use of native tags like `<h1> <h2> <h3>` etc, to handle this for you! give the title `<h1>`.

index.html
```diff
@@ -1,19 +1,121 @@
...
+          <li>
+            <img
+              src="./img/twitter-icon.svg"
+              alt="twitter bird icon"
```
JoeWeate, line 90: Good use of alt! 👍

JoeWeate (changes requested):

JoeWeate:
Good job @AmarachiOkereke! You've learned a lot about how to debug your issue, and thats the biggeest thing for a Software Developer!  A little more practice with flex and you'll be ready to go. It's a very weird concept to get used to, when I started I found this page to be very useful! I would keep it printed out everytime I coded!  I would really stress spending some time just working on understanding this - everything else, like font sizes, etc will come natural to you. [FlexBox Cheatsheet](https://css-tricks.com/snippets/css/a-guide-to-flexbox/)

css/style.css
```diff
@@ -16,4 +15,314 @@ body {
...
+  margin-top: 25px;
+}
+
+.address2 {
+  margin-top: 25px;
+}
+
+
+input {
+  padding-top: 25px;
```
JoeWeate, line 238: Use padding instead of padding-top. That way your text is centered

css/style.css
```diff
@@ -16,4 +15,314 @@ body {
...
+  border-radius: 5px;
+  margin-top: 5px;
+  border: 2px solid black;
+}
+
+ #city{
```
JoeWeate, line 244: I would replicate a lot of the logic in the `input`css into one for a `select`. Then only put "city" specific things in a city tag.

css/style.css
```diff
@@ -16,4 +15,314 @@ body {
...
+   padding-top: 12.5px;
+  border-radius: 5px;
```
JoeWeate, line 246: watch your indenting

store.html
```diff
@@ -0,0 +1,143 @@
...
+       
+       
+        <label class="address2" for="address2">Address 2 *</label>
+        <input type="text" id="address2" name="address2" required>
+
+       <div class="address-info">
+        <div class="city-postcode">
+        <label for="city">City</label>
+        <select id="city" >
+            <option value ="Select-your-city...">Select your city...</option>
```
JoeWeate, line 66: You can add cities to your drop down by adding `<option value="London">London</option>` as children to the <select/> tag. 

Additionally, if you want to have a "placeholder", one tricky way to do that with a option tags is to add 
`<option disabled="" selected="">Select City...</option>`. Don't be afraid to google thesee types of questions: [Placeholder for select tag]( https://stackoverflow.com/questions/17603055/placeholder-for-select-tag)

css/style.css
```diff
@@ -16,4 +15,314 @@ body {
...
+  font-size: xxx-large;
+
+}
+
+form{
+  width: 50%;
+  height: auto; 
+  display: flex;
+  flex-direction: column;
+  margin: 30px 15px;
```
JoeWeate, line 211: Don't use margin to center! Use flex!
The form and the image should be in a flex row together so you can center them together. Then focus on turing the form into its own container, which will be a column! (with maybe some more rows inside it).

JoeWeate (changes requested):

PR #302, opened 2021-11-18
Ldn 8 - Saman Zahedi - HTML/CSS week 1

SamanZahedi:

# Your Details

- Your Name: Saman Zahedi
- Your City: London
- Your Slack Name: Saman - Trainee

# Homework Details

- Module: HTML/CSS
- Week: One

# Notes

- What did you find easy?
Writing the HTML file
- What did you find hard?
Background image setting
Nothing for this task
- Any other notes?
I worked on and off on this assignment and I'm not sure if I could complete it in suggested four hours

@beckyf0keefe


index.html
```diff
@@ -1,19 +1,90 @@
...
+        <li class="nav-item">
+          <a class="nav-link" href="#">Blog</a>
+        </li>
+        <li class="nav-item">
+          <a class="nav-link" href="#">Help</a>
+        </li>
+        <li class="nav-item">
+          <a class="nav-link" href="#">Login</a>
+        </li>
+      </ul>
```
beckyf0keefe, line 44: Is there a tag that we can use here to make sure its clear that this is the navbar?

index.html
```diff
@@ -1,19 +1,90 @@
...
+            <div class="article-title">Boost your productivity</div>
+          </div>
+          <div class="article">
+            <img class="article-img" src="img/icon-refill.svg" alt="Gas pump" />
+            <div class="article-title">Pay as you Go</div>
+          </div>
+        </div>
+      </main>
+    </div>
+    <footer class="footer">
```
beckyf0keefe, line 78: Good use of semantic tags here

index.html
```diff
@@ -1,19 +1,90 @@
...
+      <p>Join us on</p>
+      <div class="footer-content">
+        <a href="#"><img src="img/twitter-icon.svg" alt="Twitter icon" /></a>
```
beckyf0keefe, line 81: Good to see you using alt tags here

index.html
```diff
@@ -1,19 +1,90 @@
...
+        <li class="nav-item">
+          <a class="nav-link" href="#">Login</a>
+        </li>
+      </ul>
+    </header>
+    <div class="content">
+      <main class="main">
+        <div class="center">
+          <p id="p1">Introducing karma</p>
+          <p id="p2">Bring WiFi with you, everywhere you go.</p>
```
beckyf0keefe, line 50: Are there semantic tags that would be better used here? For text that isn't a paragraph there are more specific tags that we can use, they also help define the structure of the page for screen reader users so they're important to use

css/style.css
```diff
@@ -1,19 +1,170 @@
...
+}
+
+.nav-link:hover {
+  color: rgb(255, 102, 0);
+  cursor: pointer;
+}
+
+/* main */
+
+.content p {
```
beckyf0keefe, line 66: Good use of combining tags to select specific elements

css/style.css
```diff
@@ -1,19 +1,170 @@
...
+  text-align: center;
+  font-size: 3rem;
+  font-weight: 100;
+}
+
+.center {
```
beckyf0keefe, line 72: Is this a descriptive classname for the hero container? center could be lots of different things so try and name your classes to be a little more reader friendly, imagine I am reading this without the HTML, I still want to have an idea where the class is being applied

css/style.css
```diff
@@ -1,19 +1,170 @@
...
+
+.center p {
+  text-align: center;
+  font-size: 2.5rem;
+  margin-top: auto;
+  color: white;
+  margin-bottom: 2rem;
+}
 
+#p1 {
```
beckyf0keefe, line 90: We try and avoid using ids and use classname instead so that we can make sure we can reuse the class later

css/style.css
```diff
@@ -1,19 +1,170 @@
...
+  display: flex;
+}
+
+footer img {
+  height: 20px;
+  padding: 15px;
+}
+
+.footer-content {
+  display: flex;
```
beckyf0keefe, line 155: Good to see you using flex-box

css/style.css
```diff
@@ -1,19 +1,170 @@
...
+  text-align: center;
+  font-size: 3rem;
+  font-weight: 100;
+}
+
+.center {
+  background-image: url("../img/first-background.jpg");
+  height: 40vw;
+  background-repeat: no-repeat;
+  padding-top: 15.3rem;
```
beckyf0keefe, line 76: This is fine given we weren't trying to make the design reactive but if you wanted a stretch goal: currently the text moves  out of the center of the image because of the top padding, can you find a way to make is so that the top padding change as the width changes?

css/style.css
```diff
@@ -1,19 +1,170 @@
...
+  background-size: 117%;
+  background-position-x: center;
+  padding-right: 5%;
+}
```
beckyf0keefe, line 80: Is there a way that we could change this to use flex-box to position the text in the center of the image instead of padding? doing it this way we have had to add padding and text-align elsewhere so it would probably give a neater solution with less lines of css :)

beckyf0keefe (commented):
Over all its looking good, the site was very close visually. With your css just make sure you're not using id instead of classes and in your HTML try and use more semantic tags, where could we have used header, section or h1 tags?

index.html
SamanZahedi: <nav> tag added

index.html
SamanZahedi: done, thanks

css/style.css
SamanZahedi: changed it to class="hero"

css/style.css
SamanZahedi: modified, thanks

PR #303, opened 2021-11-18
LDN8 Jude Ricketts Initial Commit HTML / CSS - Week 2

LovesPictures:
Updated Files

# Your Details

- Your Name: Jude Ricketts 
- Your City: LDN 8 
- Your Slack Name: LovesPictures

# Homework Details

- Module: H /CSS
- Week: 2

# Notes

- What did you find easy?  Styling

- What did you find hard? understanding the effects combined attributes

- What do you still not understand?  More clarity is needed on the combined effects of attributes; what to think about first, what are good approaches to take when designing sites from the ground up.  it would have been better to see this fully demonstrated, rather than the abstract approach that was taken in the session.

- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,11 +1,5 @@
 
  /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
-
-body {
-    font-family: 'Roboto', sans-serif;
-    -webkit-font-smoothing: antialiased;
-}
-
 /**
  * Add your custom styles below
  * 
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -14,6 +8,197 @@ body {
...
+     font: inherit;
+     vertical-align: baseline;
+     box-sizing: border-box;
+     font-family: "Roboto", sans-serif;
+     color: var(--grey-dark);
+ }
+
+ /*  Setion 1: navigation header  */
+ /* section 1 container */
+ .content1 {
```
Simon2828, line 50: Could you try to choose a more specific name? Ask yourself, what does `.content1` do? Naming can be tricky... 😅  some class names can be content-semantic (e.g. `.header` or `.submenu`) and others visually semantic (e.g. `.column4` or `.btn-success`) or functional (e.g. what it does...) so I think in this case, `.content1` could be renamed `.flex-container`

index.html
```diff
@@ -8,12 +8,110 @@
...
     <!-- All the images you need are in the 'img' folder -->
 
+    <title> Judith Ricketts | Karma Code  | </title> <!-- the document's title that is shown in a browser's title bar or a page's tab -->
+</head>
+
+<body>
+
+  <header>
+    <nav class="top-navigation">
+      <div class="content1">
```
Simon2828, line 22: Is this `div` necessary? You could try removing it and adding the class name on the `nav` element above... it's good to remove any `div`s you can so your HTML is as semantic as possible.

Simon2828 (commented):
Lots of excellent semantic HTML! Looks great 😁

PR #307, opened 2021-11-18
London class 8/iwona-HTML-CSS-Week2-Karma-website

IwonaKrysp:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


IwonaKrysp:
Hi @haack, 
Please see "Karma front page" commit, this is completed version, thanks.

css/style.css
```diff
@@ -1,19 +1,168 @@
...
 
- /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
+:root {
+  --grey-dark: #292b2c;
+  --grey-mid: #808080;
+  --grey-light: #e4e4e4;
+  --orange-dark: #c05326;
+  --orange-light: #f7eae4;
+  --white: #fff;
+}
```
haack, line 10: Good use of css variables (even though we haven't learnt them yet!), this helps to make your code really clean 🙌 

Did you find it useful using them?

css/style.css
```diff
@@ -1,19 +1,168 @@
...
+  background-image: url("../img/first-background.jpg");
+  background-color: var(--grey-light);
+  margin-bottom: 50px;
+  color: white;
+  text-align: center;
+  padding: 30px 10px;
+  background-size: cover;
+  background-position-x: center;
+  background-repeat: no-repeat;
+  height: 75vh;
```
haack, line 88: I think this is a bit big compared to the example image, you don't need to perfectly match the designs but it is it good practice because sometimes it is important as a web developer (although usually you won't just have an image that you have to try to match!)

haack (commented):
Great! Your code is really clean and well structured which is brilliant!

Well done for using flex box, having really clear names and good modular classes. 

If you want to practice in future, maybe try and see if you can get even closer to the designs. Keep it up!

css/style.css
IwonaKrysp: I totally love it. I first encounter it during "Intro to Digital Course". It is maybe more typing at the beginning but when code gets longer and longer, it make my life easier.

IwonaKrysp:
Hi @haack, 
This week homework is under "Karma Store" commit, thanks.

index.html
```diff
@@ -118,7 +118,9 @@ <h2 class="small_poster_title">
                 anymore!</q
               >
             </h2>
-            <button class="button">Get Karma today</button>
+            <a href="store.html"
+              ><button class="button">Get Karma today</button></a
+            >
```
IwonaKrysp, line 123: The spacing in line 121 and 122 is odd but every time I press "save" it come out like this, could that be extension fault?

css/style.css
```diff
@@ -217,35 +226,119 @@ q::after {
...
- feature
- feature
- feature 
+  input {
+    width: 100%;
+    margin-top: 10px;
+    margin-bottom: 20px;
+    padding: 10px 0 5px 5px;
+    border-color: var(--grey-light);
+    border-radius: 4px;
```
IwonaKrysp, line 269: problem with the border radius, I add 4px for all sides but top and left side are stronger then bottom and right, I'm not able to find where is this coming from? I did not declare border elsewhere.

index.html
IwonaKrysp: the same problem with spacing on line 116 to 119.

css/style.css
```diff
@@ -217,35 +226,119 @@ q::after {
...
+    background-size: cover;
+    background-repeat: no-repeat;
+  }
 
+  .form_field {
+    width: 75%;
+    margin: 20% auto;
+    text-align: left;
+    border: 0;
+  }
```
IwonaKrysp, line 252: I'm not really sure how to get proportions right, like in the case of fieldset size compering with other elements on the size, and figma is only somewhat helpful.

css/style.css
haack: This is unrelated to the border radius (you can see it's the same even if you remove the border-radius). Remember the border-radius on controls how round the corners are. The coloring of the border is determining the difference between the sides. Try using the `border` property to change that!

You can see examples here: https://www.w3schools.com/cssref/css3_pr_border-radius.asp

css/style.css
```diff
@@ -217,35 +226,119 @@ q::after {
...
+    align-items: center;
+    margin: 0;
+  }
+  .terms {
+    color: var(--orange-dark);
+    text-decoration: underline;
+  }
+  .submit_btn {
+    width: 50%;
+  }
```
haack, line 343: Rather than setting a % here (which will be determined by how big the container is), maybe try size it just by setting padding on the button? This will make sure that the button doesn't get too small or too big

store.html
```diff
@@ -58,7 +58,106 @@
...
+
+                <input class="button submit_btn" type="submit" value="Place my order" />
+              </fieldset>
+            </form>
+          </div>
+           <!--  __________________form side poster ___________________________ -->
+          <img
+            class="form_poster"
+            src="img/andrew-neel-QLqNalPe0RA-unsplash.jpg"
+            alt=" girl working at desk"
```
haack, line 157: Nit: extra space here

index.html
haack: Very weird, are you using prettier?

store.html
```diff
@@ -58,7 +58,106 @@
...
+                              />
+                              <label class="radio_label" for="orange">Karma orange</label>
+                              <input class =  "radio"
+                                type="radio"
+                                id="spacegray"
+                                value="space_gray"
+                                name="color"
+                              />
+                              <label for="gray">Space gray</label>
+                        </div>
```
haack, line 137: Nit: Some of the spacing here is inconsistent with the rest of the page. It'll make it easier to match the components and read it if the spacing consistent

haack (commented):
Great job, the form looks good and the navigation works. Well done for continuing to use really clear class names 🙌 

Also when you see a comment that starts with "Nit", it is short for "nit pick". This means it is a comment about a minor thing but we want to practice having a really good attention to detail so don't forgot them!

index.html
IwonaKrysp: Yes, prettier is enabled.

css/style.css
IwonaKrysp: My apologies, I must been half asleep when I was writing this, I meant border not border-radius. But I can see my mistake, I was relying on inherited border value and then I added color, what I should do is to declare again "border: 1px"  and then add color.

css/style.css
IwonaKrysp: I did not take that under account, thank you very much for pointing this.

PR #311, opened 2021-11-18
London class-8-Helena-Rog-HTML/CSS-week-2

hachi-ops:

# Your Details

- Your Name: Helena Rog
- Your City: London
- Your Slack Name: Helena Rog

# Homework Details

- Module: HTML/CSS
- Week: 2

# Notes

- What did you find easy?
- 
Layout/hover

- What did you find hard?

Matching colours/paddings/font size/weight

My way around github

- Any other notes?

That's it


css/style.css
```diff
@@ -16,4 +12,259 @@ body {
...
+}
+
+html, body {
+    font-family: 'Roboto', sans-serif;
+    -webkit-font-smoothing: antialiased;
+    margin: 0;
+    padding: 0;
+    box-sizing: border-box;
+    width: 100%;    
+    font-family: Roboto, sans-serif;
```
jstadnik, line 29: The font in your page is showing up as `sans-serif` for me -- I think this may be due to the duplicated `font-family` property in this section. Notice how this one (which most likely overwrites the top one, as css is ranked from bottom to top) does not put `Roboto` in quotes -- maybe that's why?

index.html
```diff
@@ -1,3 +1,7 @@
+<!-- Add your HTML markup here -->
```
jstadnik, line 1: No need to leave these comments behind! In fact, all non-relevant comments should not make it to the final PR.

index.html
```diff
@@ -10,10 +14,70 @@
...
+        <img class="logo" src="img/karma-logo.svg" alt="logo" width="35"/>
+        
+        <ul>
+            <li><a href="#" class="active">Meet Karma</a></li>
+            <li><a href="#">How It Works</a></li>
+            <li><a href="#">Store</a></li>
+            <li><a href="#">Blog</a></li>
+            <li><a href="#">Login</a></li>
+        </ul>
+        <img id="hamburger" src="img/menu-hamburger.svg" alt=""/>
```
jstadnik, line 27: I was not able to make this image appear?

index.html
```diff
@@ -10,10 +14,70 @@
...
+    </nav>
+    <header class="hero">
```
jstadnik, line 29: The header is the part at the very top -- the logo and navigation. By that definition, `hero` should never find itself in the header, really. It should be a normal section. But it is indeed the hero :)

css/style.css
```diff
@@ -16,4 +12,259 @@ body {
...
+nav ul li a:hover {
+    color: var(--orange);
+}
+
+.active {
+    color: black;
+}
+
+#hamburger {
+    display: none;
```
jstadnik, line 85: Things with display `none` should not be left behind, unless the display settings are toggled in some way to make them appear at least sometimes.

index.html
```diff
@@ -10,10 +14,70 @@
...
+            <div class="hero-text-and-button">
+                <h1>Introducing Karma</h1>
+                <p>Bring WiFi with you, everywhere you go.</p>
+                <button>Learn More</button>
+            </div>
+    </header>
+  
+    <main>
+        <section>
+            <div>
```
jstadnik, line 40: Do you actually need a `div` here? I cannot see it being targeted by any css rules.

index.html
```diff
@@ -10,10 +14,70 @@
...
+    </header>
+  
+    <main>
+        <section>
+            <div>                
+                <h2>Everyone needs a little Karma</h2>
+            </div> 
+            <div class="section-items">
+                <div class="section-item">
+                    <img src="img/icon-devices.svg" alt="" />
```
jstadnik, line 45: Images should have an alt text -- otherwise someone accessing this on a screen reader would not be able to tell what they are. For example "phone, computer and tablet".

css/style.css
```diff
@@ -16,4 +12,259 @@ body {
...
+    border-radius: 6px;
+    display: block;
+    margin: 0 auto;
+    margin-top: 80px;
+}
+
+
+/* Section */
+
+.section-items {
```
jstadnik, line 150: Good use of flexbox!

css/style.css
```diff
@@ -16,4 +12,259 @@ body {
...
+
+.section-item {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;    
+}
+
+.section-item img {
+    width: 160px;
```
jstadnik, line 166: Should not be pixels -- use relative measurements where possible (need to be careful if they're not svg files, in case the relative measurement starts to be bigger than the actual image size, resulting in blurring)

index.html
```diff
@@ -10,10 +14,70 @@
...
+                </div>            
+        </section>
+        
+    </main>
+    
+    <footer>
+        <hr/> 
+        <p class="first-p">Join us on</p>
+        <div class="footer-icons">
+            <img src="img/twitter-icon.svg" alt="twitter" width="20"/>
```
jstadnik, line 75: As discussed in class, these should be wrapped in a `div` to prevent the borded squashing the image.

jstadnik (commented):
Easy to read code, with just enough classes to make them meanigful/useful, but at the same time easily referenceable -- a pleasure to review :)

Great work on meeting your deadlines ahead of time :)

IrinShilova:
Helena, I think your Karma page looks great! I liked how you used semantic tags. And I liked the clean structure. Well done!

PR #312, opened 2021-11-19
london-class-8/shadi-safaee/karma-week-2

ShadiSafaee:

# Your Details

- Your Name: Shadi Safaee
- Your City: Winchester
- Your Slack Name: Shadi

# Homework Details

- Module: HTML & CSS
- Week: 2

# Notes

- What did you find easy? All easy

- What did you find hard? nothing

- What do you still not understand? nothing

- Any other notes? -

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -14,87 +14,102 @@ body {
  * 
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
-.topnav {
-  background-color: rgb(255, 255, 255);
-  overflow: hidden;
-  margin-right: 60px;
-}
 
+header {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  padding: 20px 5%;
+  height: 5%;
+  align-items: center;
+}
 /* nav bar */
-.topnav a {
-  float: right;
-  color: #5a5a5a;
-  text-align: center;
-  padding: 1vw 2vw;
+
+nav {
+  display: block;
+}
+.nav__links {
+  list-style: none;
+}
+.nav__links li {
+  display: inline-block;
+  padding: 0px 15px;
+}
+a {
   text-decoration: none;
-  font-size: 2vw;
-  background-color: white;
-  margin-top: 2vw;
+  color: rgb(116, 114, 114);
 }
 
 /* the color of links on hover */
-.topnav a:hover {
+.nav__links li a:hover {
   color: orangered;
 }
 
 /* Adding bold font to the active/current link */
-.topnav a.active {
+.active {
   font-weight: bold;
 }
+/* logo */
 .logo {
-  margin-left: 4vw;
-  margin-top: 2vw;
-  height: 3vw;
-  width: 2vw;
+  height: 30px;
+  width: 25px;
+  cursor: pointer;
 }
+.logo-top {
+  width: 25%;
+  padding: 10px;
+}
+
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -1,19 +1,102 @@
...
+        <div id="item">
+           <img class="icon" id="thumbnail" src="img/icon-coffee.svg" alt="coffee"/>
+           <span class="sub-title">Boost your productivity</span>
+        </div>
+        <div id="item">
+           <img class="icon" id="thumbnail" src="img/icon-refill.svg" alt="re-fill"/>
+           <span class="sub-title">Pay as you go</span>
+        </div>
+      </div>
+    </section>
```
beckyf0keefe, line 70: good use of the section semantic tag

css/style.css
```diff
@@ -15,5 +14,147 @@ body {
...
+#title-main {
+  color: rgb(70, 69, 69);
+  text-align: center;
+  margin-top: 10vw;
+}
+.icon {
+  width: 15vw;
+  height: 15vw;
+}
+#thumbnail:hover {
```
beckyf0keefe, line 103: good to see you using the :hover selector, it will be really important in the future

css/style.css
```diff
@@ -15,5 +14,147 @@ body {
...
+  border-radius: 3%;
+  width: 30%;
+  font-size: 1.5vw;
+}
+
+.article-summary {
+  text-align: center;
+}
+
+#title-main {
```
beckyf0keefe, line 94: try not to use ids in your css, if we make this a class then we can reuse elsewhere if we need to

css/style.css
```diff
@@ -15,5 +14,147 @@ body {
...
+  margin-top: 2vw;
+  height: 3vw;
+  width: 2vw;
+}
+/* styling background image & texts on the image */
+
+.big-image {
+  width: 97%;
+  margin: 0;
+}
```
beckyf0keefe, line 55: On the image that we are recreating the hero image goes all the way across the screen, a keen eye for detail is really important to make sure that we are recreating it exactly to what the image shows

css/style.css
```diff
@@ -15,5 +14,147 @@ body {
...
+.button1 {
+  background-color: orangered;
+  border: 2vw solid orangered;
+  border-radius: 3%;
+  width: 30%;
+  font-size: 1.5vw;
+}
+
+.article-summary {
```
beckyf0keefe, line 90: this is one way to do this but after what we learned in class last week, try to have a go at using flex-box instead to achieve the same effect

css/style.css
```diff
@@ -15,5 +14,147 @@ body {
...
+  height: 2px;
+  border-width: 0;
+  color: rgb(196, 196, 196);
+  background-color: rgb(196, 196, 196);
+}
 
+footer {
+  display: flex;
+  justify-content: center;
+}
```
beckyf0keefe, line 155: good use of flex-box

beckyf0keefe (commented):
Over all it looks like you're getting there with semantic tags, just make sure to use it for all the relevant content. The site visually was looking good so overall well done with the css, just make sure that you are using flexbox where appropriate and avoiding using ids as css selectors.
As a goal if you have time please change the hero image to a css background image and change the alt tags to be more descriptive.

ShadiSafaee (commented):
Thank you Becky for all your comments, I really appreciate it. All updated based on your comments.

PR #321, opened 2021-11-19
London class 8 - Hamra Khan - HTML/CSS - Week 2

HamraKhan:

# Your Details

- Your Name: Hamra Khan
- Your City: London
- Your Slack Name: Hamra Khan

# Homework Details

- Module: HTML/CSS
- Week: 2

# Notes

- What did you find easy?
I found placing the semantic elements and writing the appropriate convention html/css code wasn't too difficult. I find the more I do it the better understanding I get of the layout and how it should be.  

- What did you find hard?
Implementing flex box and minimising code in order to avoid repetitions and unnecessary code

I find I still get a bit stuck using flex box.

- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -6,17 +6,7 @@ body {
     -webkit-font-smoothing: antialiased;
 }
 
-/**
- * Add your custom styles below
- * 
- * Remember: 
- * - Be organised, use comments and separate your styles into meaningful chunks
- *    for example: General styles, Navigation styles, Hero styles, Footer etc.
- * 
- * - When using Flexbox, remember the items you want to move around need to be inside a parent 
- container set to 'display: flex'*/
-
- /* HEADER */
+/* HEADER */
 .header {
     display: flex;
 }
@@ -24,7 +14,7 @@ body {
 .header-image {
     height: 50px;
     width: 50px;
-    padding-left: 150px;
+    padding-left: 160px;
     padding-top: 15px;
 }
 
@@ -32,7 +22,7 @@ body {
 .nav-list {
     display: flex;
     list-style: none;
-    margin-left: 1020px;
+    margin-left: 900px;
 }
 
 .nav-item {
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -1,19 +1,167 @@
...
- * 
- * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
- */
+/* HEADER */
+.header {
+    display: flex;
+}
+
+.header-image {
+    height: 50px;
```
ewintram, line 14: in this case it looks fine, but a safer way to size images is to create a parent div for an img element and set height & width on the parent. Then you can set the height & width of the img element to be 100% and use the [object-fit](https://www.w3schools.com/css/css3_object-fit.asp) property to ensure the image is scaled correctly within the parent container

css/style.css
```diff
@@ -1,19 +1,167 @@
...
+    width: 50px;
+    padding-left: 160px;
+    padding-top: 15px;
+}
+
+/* NAVIGATION */
+.nav-list {
+    display: flex;
+    list-style: none;
+    margin-left: 900px;
```
ewintram, line 24: this large margin is making the <nav> disappear off my screen to the right. If you want the header-image to be left aligned and the <nav> to be right-aligned, can you think of a flex-box property you can use on the `header` class instead of this `margin-left property`?

css/style.css
```diff
@@ -1,19 +1,167 @@
...
+}
+
+.nav-item {
+    padding: 17px;
+}
+
+.nav-link:hover {
+    color: #eb6728;
```
ewintram, line 32: love the hover effect! Remember you can define CSS variables for colours so you don't have to remember what these Hex codes mean, e.g.
```
:root {
  --orange: #eb6728;
}
```

css/style.css
```diff
@@ -1,19 +1,167 @@
...
+   text-decoration: none; 
+   color: #888;
+}
+
+/* BACKGROUND IMAGE, TEXT AND BUTTON */
+.page-top-half {
+    position: relative;
+}
 
+.page-top-half .large-text {
```
ewintram, line 45: nice job separating the shared styles with the additional class `large-text` and then adding the individual styles to `h1` and `p`!

css/style.css
```diff
@@ -1,19 +1,167 @@
...
+    margin-inline-end: 0px;
+    font-size: 60px;
+    font-weight: 300;
+    color: #3c3b3b;
+}
+
+.icon-container {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-evenly;
```
ewintram, line 108: wonderful use of flex-box! By default `flex-direction` is set to `row` so you don't need to provide the property unless you're overriding where it was set by another selector

css/style.css
```diff
@@ -1,19 +1,167 @@
...
+}
+
+.icon {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
```
ewintram, line 114: nice use of flex-box again!

css/style.css
```diff
@@ -1,19 +1,167 @@
...
+    color: #3c3b3b;  
+    margin-bottom: 20px;
+    margin-top: 30px;
+}
+
+.footer-content {
+    margin-bottom: 20px;
+    display: flex;
+    flex-direction: row;
+    justify-content: center;
```
ewintram, line 145: You've got it!! Another spot on use

index.html
```diff
@@ -9,11 +9,102 @@
...
+                    </li>
+                    <li class="nav-item">
+                        <a class="nav-link" href="#">Help</a> 
+                    </li>
+                    <li class="nav-item">
+                        <a class="nav-link" href="#">Login</a> 
+                    </li>
+                </ul>
+            </nav>
+        </header>
```
ewintram, line 42: Great use of semantic elements and CSS class names

index.html
```diff
@@ -9,11 +9,102 @@
...
-</body>
+        <!-- MAIN -->
+        <main>
+            <!-- TOP HALF OF PAGE -->
+            <div class="page-top-half">
+                <img class="background-image" src="img/first-background.jpg" alt="background-image"/>
+
+                <!-- TEXT/BUTTON ON BACKGROUND IMAGE -->
+                <h1 class="large-text">Introducing Karma</h1>
+                <p class="large-text">Bring WiFi with you, everywhere you go.</p>
```
ewintram, line 52: Maybe this would be better as an `h2` element instead of `p` as it's more like a sub-heading than a paragraph

ewintram (commented):
The page looks super, well done! The HTML is neat and excellent use of semantic elements. Also your use of flex-box is great!!

PR #330, opened 2021-11-20
Karma Progress So Far

SimonLawrenceGB:
Note quite completed yet but close.

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,9 +1,11 @@
-
- /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
+/* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
+:root {
+  --orange-karma: #f26639;
+}
 
 body {
-    font-family: 'Roboto', sans-serif;
-    -webkit-font-smoothing: antialiased;
+  font-family: "Roboto", sans-serif;
+  -webkit-font-smoothing: antialiased;
 }
 
 /**
@@ -16,4 +18,96 @@ body {
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
 
+/*Header*/
 
+header {
+  display: flex;
+  padding: 0 5rem;
+  justify-content: space-between;
+}
+
+/*Karma Logo*/
+
+header img {
+  height: 3rem;
+  width: 3rem;
+  margin-top: 10px;
+}
+
+/*Navigation*/
+
+ul {
+  display: flex;
+  list-style: none;
+  justify-content: flex-end;
+}
+
+li {
+  padding: 0.5rem;
+}
+
+a {
+  color: gray;
+  text-decoration: none;
+}
+
+a:hover {
+  color: var(--orange-karma);
+}
+
+/*Intro Image*/
+
+.intro {
+  display: flex;
+  background-image: url("../img/first-background.jpg");
+  padding: 15.5rem 0;
+  color: white;
+  flex-direction: column;
+  align-items: center;
+  background-repeat: no-repeat;
+}
+
+/*Intro Text*/
+
+h1,
+h2,
+h3,
+.intro p {
+  font-weight: lighter;
+}
+
+h1 {
+  font-size: 2.5rem;
+  margin: 0;
+}
+
+.intro p {
+  font-size: 1.4rem;
+}
+
+button {
+  padding: 0.75rem 1.25rem;
+  background-color: #f26639;
+  border: none;
+  border-radius: 5px;
+  font-size: 0.75rem;
+  margin-top: 2rem;
+}
+
+/*Icon Image*/
+
+.icon_images {
+  display: flex;
+  height: 10rem;
+  width: 10rem;
+}
+
+/*Icon Text*/
+
+/*Social Media*/
+
+.social_media_instagram {
+  display: flex;
+  height: 10rem;
+  width: 10rem;
+}
--- a/index.html
+++ b/index.html
@@ -1,19 +1,94 @@
 <!DOCTYPE html>
 <html lang="en">
-<head>
-    <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+  <head>
+    <meta charset="UTF-8" />
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1, shrink-to-fit=no"
+    />
     <title>Karma</title>
-    <link href='https://fonts.googleapis.com/css?family=Roboto:400,500,300' rel='stylesheet' type='text/css'>
-    <link rel="stylesheet" href="css/normalize.css">
-    <link rel="stylesheet" href="css/style.css">
-    <link rel="shortcut icon" type="image/x-icon"  href="favicon.ico">
-</head>
-<body>
+    <link
+      href="https://fonts.googleapis.com/css?family=Roboto:400,500,300"
+      rel="stylesheet"
+      type="text/css"
+    />
+    <link rel="stylesheet" href="css/normalize.css" />
+    <link rel="stylesheet" href="css/style.css" />
+    <link rel="shortcut icon" type="image/x-icon" href="favicon.ico" />
+  </head>
+
+  <body>
+    <header>
+      <img src="img/karma-logo.svg" alt="Karma Logo" />
+      <nav>
+        <ul>
+          <li>
+            <a href="#"><b>Meet Karma</b></a>
+          </li>
+          <li>
+            <a href="#">How it Works</a>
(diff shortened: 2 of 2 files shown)


css/style.css
```diff
@@ -16,4 +19,144 @@ body {
...
+
+.social_media_images {
+  display: flex;
+  height: 1.5rem;
+  padding: 0.5rem;
+  margin: 0rem 0.5rem;
+  border: 2px solid var(--cream-karma);
+  border-radius: 30px;
+}
+
```
jstadnik, line 156: As you might have noticed, the images do not show upon the page correctly -- this is because while the div they are in is sized, the images themselves are not. What you want, is to constrain them to follow the size of the div, for example:

```css
.social_media_images img {
  height: 100%;
}
```

Just make sure you only fix one dimension, not both! This will allow the image to retain the original ratios and not get distorted.

css/style.css
```diff
@@ -16,4 +19,144 @@ body {
...
+/*Social Media Style*/
+
+.social_media {
+  display: flex;
+  justify-content: center;
+}
+
+.social_media_images {
+  display: flex;
```
jstadnik, line 149: `display: flex` should only be used on the parent element, not the items. Given you're not styling the items (`img`s) anywhere, there's only one `img` per `.social_media_images`, and there's no other flex properties used in the parent css, I assume this is a bug?

index.html
```diff
@@ -1,19 +1,96 @@
...
+        </div>
+      </section>
+
+      <hr />
+    </main>
+
+    <footer>
+      <h3><b>Join us on</b></h3>
+      <section class="social_media">
+        <a class="social_media_images" href="#">
```
jstadnik, line 77: Points for extra effort in making these links :)

index.html
```diff
@@ -1,19 +1,96 @@
...
+          <img src="img/instagram-icon.svg" />
+        </a>
+      </section>
+      <div>
+        <p class="karma_mobility">&#169;Karma Mobility, Inc.</p>
+      </div>
+    </footer>
 
     <!-- Add your HTML markup here -->
```
jstadnik, line 92: Leftover comments.

index.html
```diff
@@ -1,19 +1,96 @@
...
+            <a href="#">Help</a>
+          </li>
+          <li>
+            <a href="#">Login</a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+
+    <main>
```
jstadnik, line 47: Good use of semantic html tags :+1:

index.html
```diff
@@ -1,19 +1,96 @@
...
+      </section>
+      <section>
+        <h2>Everyone needs a little Karma</h2>
+        <div class="karma_icon_style">
+          <div class="karma_icon_direction">
+            <img class="karma_icon_images" src="img/icon-devices.svg" />
+            <p>Internet for all devices</p>
+          </div>
+          <div class="karma_icon_direction">
+            <img class="karma_icon_images" src="img/icon-coffee.svg" />
```
jstadnik, line 61: You should try to have alt text on all your images, to help those reading using a screen reader.

css/style.css
```diff
@@ -16,4 +19,144 @@ body {
...
+
+.intro p {
+  font-size: 1.5rem;
+}
 
+/*Intro Button*/
+
+button {
+  padding: 0.75rem 1.25rem;
+  background-color: #f26639;
```
jstadnik, line 98: Missed this one var, but in general, good work on adding variables to your CSS :tada:

jstadnik (commented):
Easy to review, thanks to a good use of semantic tags and class names :+1: 

With a small correction for image size, the page renders nicely and very close to the picture (kind of difficult to tell for sure with a picture, honestly).

Great work! :)

jstadnik:
Also, one thing -- I think you're supposed to fill in the answers on your PR post, rather than in the template.

css/style.css
SimonLawrenceGB: Hi Joanna, I added the lines to my CSS file but I couldn't see any difference. Are you referring to just the social media images or all the images?

css/style.css
SimonLawrenceGB: The unnecessary display: flex has been removed.

index.html
SimonLawrenceGB: Removed.

index.html
SimonLawrenceGB: Added alt tags to images.

css/style.css
SimonLawrenceGB: Corrected missing var.

css/style.css
SimonLawrenceGB: Never mind, I understand now. I got confused because I removed display flex from social_media_images. It had the same effect as adding the lines you suggested.

jstadnik: Oh, I think I made this mistake the other way around! I added the additional img styling first, and then simply checked removing `display: flex` didn't break anything...

index.html
```diff
@@ -72,13 +72,32 @@ <h2>Everyone needs a little Karma</h2>
...
+          <img
+            class="karma_feature_image"
+            src="additional/feature-img.png"
+            alt="Image of women sitting down and smiling whilst using a tablet"
+          />
+        </div>
+        <div class="wherever_i_am">
+          <p>
+            <em>
+              <i>"</i>Wherever I am, I just don't worry about my connection
```
jstadnik, line 94: There's an element called `q` that can be used for the quotes.

SimonLawrenceGB: Yes, but I only want the quotation marks to turn orange and not the whole quote. I tried doing this with q but couldn't figure it out.

index.html
```diff
@@ -72,13 +72,32 @@ <h2>Everyone needs a little Karma</h2>
           <div class="karma_icon_direction">
             <img
               class="karma_icon_images"
+              style="padding-left: 1rem"
```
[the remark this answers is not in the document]
SimonLawrenceGB: Corrected.

index.html
SimonLawrenceGB: Figured it out.

css/style.css
```diff
@@ -16,4 +22,202 @@ body {
...
+.karma_icon_images {
+  display: flex;
+  height: 50%;
+  margin: auto;
+  padding: 0rem 0.5rem;
+}
+
+/*Karma Feature*/
+
+.karma_feature {
```
teniolao, line 142: Good work simon

teniolao (changes requested):
good semantics layout. just need a few corrections here and there and you are good.

store.html
```diff
@@ -0,0 +1,146 @@
...
+          <fieldset>
+            <label for="address2">Address 2</label>
+            <input type="text" id="address2" name="address2" />
+          </fieldset>
+          <fieldset class="label_input">
+            <div>
+              <label for="city">City *</label>
+              <select id="city" name="city" required>
+                <option disabled hidden selected>Select your city...</option>
+                <option>London</option>
```
jstadnik, line 77: Each selectable option should have a value associated with it (this is so that the backend can know what was selected).

store.html
```diff
@@ -0,0 +1,146 @@
...
+            <div>
+              <label for="postcode">Postcode *</label>
+              <input type="text" id="postcode" name="postcode" required />
+            </div>
+          </fieldset>
+          <fieldset>
+            <p>Select a colour</p>
+            <div>
+              <label for="karma_orange"></label>
+              <input
```
jstadnik, line 89: Again, each input of type `radio` needs to have a value associated with it.

store.html
```diff
@@ -0,0 +1,146 @@
...
+                type="radio"
+                id="space_grey"
+                name="karma_orange-space_grey"
+              />
+              Space Grey
+            </div>
+          </fieldset>
+          <fieldset>
+            <label>
```
jstadnik, line 107: Instead of having the input inside the label, you should have the label and the input separately, like you did for the other inputs. You can always check for the best practice / way to do things on the mdn website: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox

store.html
```diff
@@ -0,0 +1,146 @@
...
+              <input type="checkbox" /> By placing your order you agree to
+              Karma's Terms and Conditions. *
+            </label>
+          </fieldset>
+          <fieldset>
+            <button type="submit">Place my order</button>
```
jstadnik, line 113: This should be an input, not button. Again, you can always google "html submit" -- the mdn resource comes up as the first result for me.

jstadnik (commented):
Good matching of `for` attributes and ids :)

PR #335, opened 2021-11-23
Karma coursework-HTML/CSS-week2

MonMon20:

# Your Details

- Your Name: Monica Amanfo
- Your City: London
- Your Slack Name: Monica A

# Homework Details

- Module: HTML/CSS-week2
- Week: 2

# Notes

- What did you find easy?
Using the hero image and putting text on top
- What did you find hard?
Positioning of the logo
How to resize the logo properly 
- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,9 +1,8 @@
-
- /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
+/* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
 
 body {
-    font-family: 'Roboto', sans-serif;
-    -webkit-font-smoothing: antialiased;
+  font-family: "Roboto", sans-serif;
+  -webkit-font-smoothing: antialiased;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -1,19 +1,71 @@
...
+    />
     <title>Karma</title>
-    <link href='https://fonts.googleapis.com/css?family=Roboto:400,500,300' rel='stylesheet' type='text/css'>
-    <link rel="stylesheet" href="css/normalize.css">
-    <link rel="stylesheet" href="css/style.css">
-    <link rel="shortcut icon" type="image/x-icon"  href="favicon.ico">
-</head>
-<body>
-
+    <link
```
ellietms, line 10: Hi @MonMon20 
Well done for using:

1)semantic HTML
2)HTML nesting and indentation 
3)your page looks great 🥳 🎊

what you need to improve: 👩🏼‍💻

1) you did not use classes and you just used the tag names in your CSS file, please change all of them and use classes instead.
2)you did not use flex-box, I just saw one display: flex in ul tag and one of the objectives of this homework was for you to practice more about classes and flex-box.
3) you used width and height for images, instead of doing this you have to use a div (wrapper) and wrap your images with that div and give each of the img tags a width depending on the width of the div tag.
4) you did not use classes for your parent elements.
5) in your HTML file, for some of the elements you wrote some classes but in your CSS file you do not have any class with that name(you did not create any classes) which will make confusion, it is better whenever you want to create a class, just choose  a meaningful name for your class and create it in your CSS file at the same time ( to prevent to forget it somehow 😉 )

index.html
```diff
@@ -1,19 +1,71 @@
...
+    />
+    <link rel="stylesheet" href="css/normalize.css" />
+    <link rel="stylesheet" href="css/style.css" />
+    <link rel="shortcut icon" type="image/x-icon" href="favicon.ico" />
+  </head>
+  <body>
+    <header>
+      <img src="img/karma-logo.svg" alt="Karma logo"/>
+      <nav style="height: 71px; bottom: 107px">
+        <ul>
```
ellietms, line 23: @MonMon20  you should not use one line CSS like this anymore, you should use classes and use that class here.

css/style.css
```diff
@@ -16,4 +15,125 @@ body {
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
 
+header {
```
ellietms, line 18: instead of using the tag names in your CSS, it is better to create some classes and use those classes because when you have a bigger website, if you write your CSS in this way it will affect all of the tags on different pages.

css/style.css
```diff
@@ -16,4 +15,125 @@ body {
...
+  margin-top: 0;
+  position: relative;
+  z-index: 5;
+  height: 71;
+  bottom: 107;
+}
+
+ul {
+  display: flex;
+  list-style: none;
```
ellietms, line 33: it is okay that you used "display: flex"  for `ul` here because you just have one "ul", but generally, you should always create classes in your CSS file and  use `display: flex`  for the parent element and align its child element with other flex-box features

css/style.css
```diff
@@ -16,4 +15,125 @@ body {
...
+  margin-top: 120px;
+}
+
+li:hover {
+  color: rgba(224, 100, 58, 255);
+}
+
+img {
+  height: 40px;
+  top: 9px;
```
ellietms, line 54: @MonMon20  please add a wrapper (div) for your img tags and give that div a class with width:100% and give your img tags a class with width:40%

css/style.css
```diff
@@ -16,4 +15,125 @@ body {
...
+  font-size: 50px;
+  margin-top: 70px;
+}
+
+h2 {
+  font-weight: 80;
+  font-size: 30px;
+}
+main button {
+  background-color: rgba(224, 100, 58, 255);
```
ellietms, line 86: Use classes, do not use tag names like this

css/style.css
```diff
@@ -16,4 +15,125 @@ body {
...
+  border: 0;
+  border-top: 1px solid lightgray;
+}
+
+footer {
+  text-align: center;
+}
+
+footer img {
+  width: 20px;
```
ellietms, line 128: give your footer a class with width 100% and give your images inside of the footer a class with width:40%

index.html
```diff
@@ -1,19 +1,71 @@
...
+      </section>
+    </main>
+    <hr />
+    <footer>
+      <section>
+        <p>Join us on</p>
+        <img src="/img/twitter-icon.svg" alt="twitter icon"class="links">
+        <img src="/img/facebook-icon.svg" alt="facebook icon"class="links">
+        <img src="img/instagram-icon.svg" alt="instagram icon"class="links">
+        <p class="copyright">© Karma Mobility, Inc.</p>
```
ellietms, line 64: @MonMon20  I do not see any links class in your CSS file.

index.html
```diff
@@ -1,19 +1,71 @@
...
+    <main>
+      <section class="main-section">
+        <h1>Introducing Karma</h1>
+        <h2>Bring WiFi with you, everywhere you go.</h2>
+        <button>Learn more</button>
+      </section>
+      <section class="lower-section">
+        <h2>Everyone needs a little Karma.</h2>
+        <div class="bottom_images">
+          <figure>
```
ellietms, line 42: I do not see any middle-images classes in your CSS file.

index.html
```diff
@@ -1,19 +1,71 @@
...
+            <img src="/img/icon-devices.svg" alt="devices icon" class="middle-images" />
+            <figcaption>Internet for all devices</figcaption>
```
ellietms, line 44: for using the class names, it is better to use some meaningful names which make sense when someone is going through your code, for example, you used a class name "bottom-images" for the div and then for each img tags you used "middle-images" (which does not exist), so it is better to choose 2 different meaningful name which are related to each other, for example for the div you can use class = images-container" and for img tags social-media-images 🙂   Does this make sense?

ellietms (changes requested):

PR #360, opened 2022-06-13
ZA2-Chezlin Benson-Week3-Karma

chezlinbenson:
Completed the HTML and CSS for the Karma webpage/ desktop design.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -4,6 +4,47 @@
 body {
     font-family: 'Roboto', sans-serif;
     -webkit-font-smoothing: antialiased;
+    margin: 0;
+    height: 100%vh;
+    display: grid;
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -4,8 +4,154 @@
...
+
+main {
+    grid-area: main;
+    
+}
+
+.articles {
+    display: flex;
+    justify-content: center;
+    gap: 150px;
```
AlexJora, line 103: wouldn't be safer to use rem instead of px?

AlexJora (changes requested):
I suggest you use em or rem instead of px.

Xolani-M:
Could you please review my code.

Thanks.

PR #364, opened 2022-06-15
ZA2-Kabelo Letlala-HTML/CSS-Module-project, week 3

M3ntalist:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -153,16 +153,18 @@ body {
     height: 20%;
     padding-top: 1rem; 
     font-size: 1rem;
+    justify-content: space-between;
   }
   
   .icon {
     width: 20px;
     padding: 20px;
     font-size: 30px;
+    font-display: var(rgb(226, 223, 223));
     width: 30px;
     text-align: center;
     text-decoration: none;
-    border-radius: 50%;
+    border-radius: 50%; 
     border: solid 2px rgb(204, 193, 193);
   }
   
--- a/css/style.css
+++ b/css/style.css
@@ -17,3 +17,164 @@ body {
  */
 
 
+* {
+    margin: 0;
+    padding: 0;
+    /* background-color: white; */
+  }
+  
+  body {
+    width: 100%;
+  }
+  
+  
+  li,
+  a {
+    color: gray;
+    text-decoration: none;
+  }
+  
+  header {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    padding: 20px 10%;
+  }
+  
+  .logo {
+    max-height: 40px;
+    cursor: pointer;
+  }
+  
+  .nav__links {
+    list-style: none;
+  }
+  
+  .nav__links li {
+    display: inline-block;
+    padding: 0px 20px;
+  }
+  
+  .image__container {
+    background-image: url(../img/first-background.jpg);
+    background-position: center;
+    background-repeat: no-repeat;
+    background-size: cover;
+    width: 100%;
+    height: 500px;
+    color: white;
+    font-family: "Roboto", sans-serif;
+    padding-top: 335px;
+  }
+  
+  .inner__container {
+    text-align: center;
+    position: absolute;
+    width: 100%;
+    height: 450px;
+  }
+  
+  h1 {
+    font-size: 5rem;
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -14,6 +14,84 @@
...
+    <p>Bring WiFi with you, everywhere you go.</p>
+    <a class="btn" href="#">Learn More</a>
+    </div>
+    </div>
+    </section>
+    <section>
+    <div>
+    <h2>Everyone needs a little Karma</h2>
+    <div class="image__center">
+    <div>
```
Mathias02, line 65: Buddy, I am trying to figure out what's this div all about here since it's empty, please review this area,

Mathias02: Oh I see, I think it could be better if you had articles than keeping divs

M3ntalist: Hi Mathias, let me check it out bro. Thanks man

index.html
```diff
@@ -59,13 +59,13 @@ <h1>Introducing Karma</h1>
     </div>
     </section>
     <section>
-    <div>
+    <article>
```
Mathias02, line 62: I would prefer you to keep articles than mixing them with divs and the way you mixed them makes things furthermore complicated,

Mathias02 (changes requested):
Review my Karma project, please

index.html
M3ntalist: I will use div for now, as I don't have enough time to re do my work now.

Mathias02: ok cool, maybe later

index.html
chandredewet: Hi Kabelo,

Agree with Matthias a better suggestion to use articles.
From MDN [article](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article) definition suggests selfcontained content and in this case multiple selfcontained content. Of course you can keep the divs but use in future. A snap of MDN articles in action,

image

index.html
```diff
@@ -14,6 +14,88 @@
...
+</div>
+    </section>
+    </main>
 
+    <br />
+    <hr />
+    <!-- Footer -->
+    <footer>
+    <div>
+    <h4>Join us on</h4>
```
chandredewet, line 90: The use of h4 here is incorrect.

If this is a heading then you should use h2, as headings are about hierarchy not size, and the only other heading you have on your page is h1, so the next heading should be h2 and not jump to h4, and then assign in css the font size you wish for the h2.

If this is not a heading and you used h4 to get the sizing you want then you should just have used a `<p>` tag and then allocated the font size you want for it.

index.html
```diff
@@ -14,6 +14,88 @@
...
+    <div class="inner__footer">
+    <img class="icon" src="img/twitter-icon.svg" alt="Devices image" />
+    <img class="icon" src="img/facebook-icon.svg" alt="Devices image" />
+    <img class="icon" src="img/instagram-icon.svg" alt="Devices image" />
+    </div>
+    <p>&copy;Karma Mobility. Inc.</p>
```
chandredewet, line 96: This is not a issue, just wanted to commend you on the use of &copy, didn't do mine like this but now I know :)

css/style.css
```diff
@@ -16,4 +16,174 @@ body {
...
+    padding: 0;
+    /* background-color: white; */
+  }
+  
+  body {
+    width: 100%;
+  }
+  
+  
+  li,
```
chandredewet, line 31: I don't think its necessary to style both li and a... as its only the a that needs to have the colour and text-decoration none.

Unless you meant that u want to target all anchor tags within list items only then it should have been `li a { `.

chandredewet (commented):
Good Day Kabelo,

Great work, I have added some comments along the way, but I can't find much fault. Glad to see you added a media query, I did not think responsiveness is required for level 1 but does make the website instantly better on all screens.  

I don't think you have much to change but just want to reiterate the use of indentation, especially in html, as that non closing div would have been easier picked up, as well as for someone else to see what is happening in your code.

Otherwise well done. You definitely are winning in this.

PR #385, opened 2022-06-28
WM4 - Azin Yadegari - HTML / CSS - Week 1

AzinYad:

# Your Details

- Your Name: Azin
- Your City: Birmingham / WestMidlands
- Your Slack Name: AzinYad

# Homework Details

- Module:HTML / CSS
- Week: 1

# Notes

- What did you find easy?
writing basic structure with html 

- What did you find hard?
It was not hard ,I just need enough practice to build some muscle memory , as I still forgot something I already knew 

There was this unwanted gap like a margin between header and hero section , I thought it must be some automatic reserved margin or padding for some elements ,tried like ten different ways and none solved the issue , so i set a border for header and hero to see exactly what causes this , and strangely it disappears as soon as I defined a border for hero !! so I still don't now the reason and the best way to solve it.

- Any other notes?
-I think it would be more precise if we could have a Figma file ,to make a code for  it ,instead of a png picture, this way we don't need to guess  colors and font size and paddings... and the assignment can be done precisely.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,19 +1,143 @@
 
  /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
 
+
+
 body {
+
     font-family: 'Roboto', sans-serif;
     -webkit-font-smoothing: antialiased;
 }
+/* Header */
 
-/**
- * Add your custom styles below
- * 
- * Remember: 
- * - Be organised, use comments and separate your styles into meaningful chunks
- *    for example: General styles, Navigation styles, Hero styles, Footer etc.
- * 
- * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
- */
+.header {
+    padding: 1rem;
+    display: flex;
+    justify-content:space-around;
+    align-items: center;
+}
+
+.header-logo {
+height: 70px;
+width: 70px;
+}
+
+ /* navigation */
+.nav-list {
+    display: flex;
+    align-items: center;
+    list-style: none;
+}
+.nav-item{
+padding: 1rem 1rem;
+margin: 0.3rem;
+
+}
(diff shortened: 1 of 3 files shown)


Nikolov87:
Generally look very well. Very good structured and easy to read. Just in store.html - in main tag. Is not wrong, but instead div you can use article/section like in your index.html. 

Kind regards,
Petar

css/style.css
```diff
@@ -1,19 +1,200 @@
...
+    background-image: url("../img/first-background.jpg");
+    background-repeat:no-repeat ;
+    height:90vh;
+    background-size: cover;
+    color: white;
+    text-align: center;
+    border: 1px solid lightgray;
+}
+.hero h1{
+    padding-top: 18.5rem;
```
simeonbikov, line 66: I really like your job. The design is very similar to the sketch.
But to centralise the elements you use - padding-top. In your screen it should looks good but in mine for example in hero section the title and button go beyond the hero section.
I can suggest to use a feature:
wrap all elements in some <div> : 

        <section class="hero" >
            <div class="hero-container">
                <h1>Introducing Karma</h1>
                <p>Bring WiFi with you,everywhere you go.</p>
                <button class="learn-btn">Learn More</button>
            </div>
        </section>

And centralize it by using transform:

add  in hero:
.hero {
    position: relative;
}

.hero-container {
    position: absolute;
    top: 20%;
    left: 50%;
    transform: translate(-50%, -50%);
    text-align: center;
}

Or you can use Flex for that.

simeonbikov: In your order-form you can also correct the position of some fields. I used flexbox for this.

simeonbikov: but overall - good job

PR #400, opened 2022-06-30
West Midlands 4 - Alex Phillip - HTML/CSS - Week 1

Alex-Phillip:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


12shussein:
leaving a review so far excellent work to improve i would add the additional required detail for example "Your Details"

Alex-Phillip:
Thank you for your review 12shussein.

index.html
```diff
@@ -1,19 +1,123 @@
...
+    />
     <title>Karma</title>
-    <link href='https://fonts.googleapis.com/css?family=Roboto:400,500,300' rel='stylesheet' type='text/css'>
-    <link rel="stylesheet" href="css/normalize.css">
-    <link rel="stylesheet" href="css/style.css">
-    <link rel="shortcut icon" type="image/x-icon"  href="favicon.ico">
-</head>
-<body>
+    <link
+      href="https://fonts.googleapis.com/css?family=Roboto:400,500,300"
```
JadeJones2022, line 11: These three lines should all be on one line for your link. Your code is very well written and easy to read.

store.html
```diff
@@ -0,0 +1,123 @@
...
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1, shrink-to-fit=no"
+    />
+    <title>Karma</title>
+    <link
+      href="https://fonts.googleapis.com/css?family=Roboto:400,500,300"
```
JadeJones2022, line 11: Lines 11 to 13 should be on one line. Well written code that is easy to read

Alex-Phillip: Thanks Jade. Didn't clock that at all, merci!

index.html
```diff
@@ -1,19 +1,123 @@
...
+        <article class="section1-article">
+          <h1 class="section1-title">Introducing Karma</h1>
+          <h3 class="section1-text">Bring WiFi with you, everywhere you go.</h3>
+          <button class="section1-button">Learn More</button>
+        </article>
+      </section>
+      <section class="section2">
+        <h1 class="section2-title">Everyone needs a little Karma.</h1>
+        <div class="section2-articles">
+          <article class="section2-article1">
```
SaadiaELF, line 46: I cannot see you using this class anywhere for styling. Is it necessary?
Wouldn't be better for to you the same class for the same elements (that's why we use classes instead of Ids)?

css/style.css
```diff
@@ -16,4 +16,278 @@ body {
...
+.section1-title {
+    font-weight: lighter;
+}
+
+.section1-text {
+    font-weight: lighter;
+    margin-bottom: 75px;
+}
+
+.section1-button {
```
SaadiaELF, line 89: I think here as well you can use the same class for all buttons and add an id or class for a different style.
exp : <button class="orange--button" id="section3-button" >
`.orange---button :{
  background-color: coral;
    border: none;
    border-radius: 5px;
    font-size: 80%;
    padding: 20px 40px;
}
#section-button:{ 
color: white;
    text-decoration: none;}`

css/style.css
```diff
@@ -16,4 +16,278 @@ body {
...
+#city {
+    width: 300px;
+}
+
+#space-grey {
+    margin-left: 50px;
+}
+
+#checkbox-text {
+    /* font-size: 23px; */
```
SaadiaELF, line 229: Better to clean up code before the last push :D

SaadiaELF (commented):
Great work so far! Keep learning and keep improving. You'll be amazing in no time :)

dewayne51 (commented):
Great effort mate!

index.html
Alex-Phillip: Hi Saadia, you're right, I don't use the class "section2-article1" anywhere in the CSS. I am just in the habit of putting classes on most elements, even if they don't require them. Thanks for pointing it out, I will try to remember it in future. Thanks.

css/style.css
Alex-Phillip: Another bad habit, I rarely complete all the html before I start styling. I think this is what leads to what you've pointed out. Thanks again for pointing it out, it's great to get feedback.

css/style.css
Alex-Phillip: :D Yes, very true.

Alex-Phillip:
Cheers Dewayne.

PR #439, opened 2022-07-24
NW5 - Haroon Ali - HTML & CSS - Week 1

haroon-ali-dev:

# Your Details

- Your Name: Haroon Ali
- Your City: Manchester
- Your Slack Name: Haroon Ali

# Homework Details

- Module: HTML & CSS
- Week: Week 1

# Notes

- What did you find easy?
creating the hero section

- What did you find hard?
creating the nav bar

n/a

- Any other notes?
n/a

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,16 +1,11 @@
-html {
-    height: 100%;
-}
-
 body {
     font-family: 'Roboto', sans-serif;
     -webkit-font-smoothing: antialiased;
-    height: 100%;
 }
 
 nav {
     display: flex;
-    padding: 10px 150px;
+    padding: 20px 150px;
 }
 
 .nav-section {
@@ -28,7 +23,7 @@ nav {
 }
 
 .nav-link {
-    font-size: 15px;
+    font-size: 20px;
     font-weight: 500;
     color: gray;
     text-decoration: none;
@@ -42,12 +37,8 @@ nav {
     color: #52555B;
 }
 
-main {
-    height: 100%;
-}
-
 .hero {
-    height: 90%;
+    height: 850px;
     background-image: url('../img/first-background.jpg');
     background-size: cover;
     color: white;
@@ -77,4 +68,32 @@ main {
     padding: 20px 50px;
     border-radius: 10px;
     margin-top: 20px;
+}
+
+.section-info {
+    text-align: center;
+    font-weight: 300;
+    padding: 50px 0px;
+}
+
+.section-info-title {
+    font-size: 50px;
+    font-weight: inherit;
+    margin-bottom: 70px;
+}
+
+.cards-container {
+    display: flex;
+    justify-content: space-between;
+    width: 60%;
+    margin: 0 auto;
+}
+
+.card img {
+    width: 150px;
+}
+
+.card p {
+    font-size: 30px;
+    font-weight: 400;
 }
--- a/css/style.css
+++ b/css/style.css
@@ -6,14 +6,36 @@ body {
     -webkit-font-smoothing: antialiased;
 }
 
-/**
- * Add your custom styles below
- * 
- * Remember: 
- * - Be organised, use comments and separate your styles into meaningful chunks
- *    for example: General styles, Navigation styles, Hero styles, Footer etc.
- * 
- * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
- */
+nav {
+    display: flex;
+    padding: 10px 150px;
+}
 
+.nav-section {
+    width: 100%;
+}
 
+.nav-section-second {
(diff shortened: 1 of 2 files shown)


AhmedMohamed809:
That is great job Haroon .

store.html
```diff
@@ -0,0 +1,163 @@
...
+                            <input id="grey" type="radio" name="color" value="grey" required />
+                            <label for="grey">Space Grey</label>
+                        </fieldset>
+                    </div>
+                    <div class="form-c-terms">
+                        <input class="form-inline" id="terms" type="checkbox" name="terms" value="agree" required />
+                        <label class="form-inline" for="terms">By placing your order you agree to Karma's <a href="#">Terms and Conditions.</a> *</label>
+                    </div>
+                    <div class="form-c-submit">
+                        <button class="btn btn-bg">Place my order</button>
```
khmdagal, line 135: button should have type="submit" attribute, so it could be sent.

store.html
```diff
@@ -0,0 +1,163 @@
...
+                        <fieldset>
+                            <legend>Select a colour</legend>
+                            <input id="orange" type="radio" name="color" value="orange" required />
+                            <label for="orange">Karma orange</label>
+                            <input id="grey" type="radio" name="color" value="grey" required />
+                            <label for="grey">Space Grey</label>
+                        </fieldset>
+                    </div>
+                    <div class="form-c-terms">
+                        <input class="form-inline" id="terms" type="checkbox" name="terms" value="agree" required />
```
khmdagal, line 131: The check box should have (checked) attribute instead of required.

store.html
```diff
@@ -0,0 +1,163 @@
...
+                        <label class="form-inline" for="terms">By placing your order you agree to Karma's <a href="#">Terms and Conditions.</a> *</label>
+                    </div>
+                    <div class="form-c-submit">
+                        <button class="btn btn-bg">Place my order</button>
+                    </div>
+                </form>
+            </div>
+            <div class="order-section-2">
```
khmdagal, line 139: This is and empty and active section. either you comment out or delete, so to help the performance of your code.

store.html
haroon-ali-dev: checked attribute cannot be used. user must first select the checkbox after reading terms and conditions.
required attribute must be used for data validation. the user must check the checkbox before submitting the form.

store.html
haroon-ali-dev: I will make that change.

store.html
haroon-ali-dev: this section is the second cell in the second column in the main grid which is showing the image of the woman. it has a background image applied to it. it cannot be removed.

css/style.css
```diff
@@ -1,19 +1,465 @@
...
+
+.active {
+    color: #52555B;
+}
+
+/*hero*/
+
+.hero {
+    height: 850px;
+    background-image: url('../img/first-background.jpg');
```
khmdagal, line 66: The background image should be half of the screen. You can review the picture named "homepage-desktop.png".

css/style.css
```diff
@@ -1,19 +1,465 @@
...
+nav {
+    display: flex;
+    padding: 20px 150px;
+    background-color: #C4C4C41A;
+}
+
+.nav-section {
+    width: 100%;
+}
+
```
khmdagal, line 36: Your nav section should have good contrast in line with the WCAG which means Web Content Accessibility Guidelines.

image

css/style.css
```diff
@@ -1,19 +1,465 @@
...
+    background-size: 100%;
+    background-position-y: 50%;
+}
+
+.hr-order {
+    margin-top: 0px;
+}
+
+/*footer*/
+
```
khmdagal, line 435: In the footer part this is color contrast issue. and that needs to be worked on.

store.html
khmdagal: You can do that, but my view is based on the sample project that we should follow and build similar to that sample. but you can choose that, which is not that much difference.

store.html
khmdagal: I got that. in this way it look starched in width wise, you can use img element and nicely style it grid. But always look the sample.

store.html
haroon-ali-dev: we must follow design sample with critical thinking. we must think about proper data validation. we must use the required attribute because the user must select the checkbox before submitting because the device cannot be purchased without accepting terms.

we cannot use the checked attribute because the user must select the checkbox themselves. not doing so will be against rules and regulations regarding purchasing the product. we must use critical thinking when following the design. the user must select the checkbox after reading the terms. it is against purchasing rules and regulations for the user to not read the terms and conditions and then not select the checkbox and have it already selected.

PR #444, opened 2022-07-27
Leeds NW5 - Levi Udeh - HTML/CSS - Week 1

levilu800b:

# Your Details

- Your Name: Levi Udeh
- Your City: Leeds
- Your Slack Name: Levi

# Homework Details

- Module: HTML/CSS
- Week: 1

# Notes

- What did you find easy? That the design has already been made so I know what to work on each time. And the images, logo and icons were also added to the workspace, that's great.

- What did you find hard? Time

- What do you still not understand? For this nothing

- Any other notes? Thanks for making the work much easier for us rather than searching through the internet for the images and icons.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,10 +1,9 @@
+/* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
 
- /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
-
-body {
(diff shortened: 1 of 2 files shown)


Meronggk (changes requested):
you can add the image on the html part instead of the css

PR #450, opened 2022-07-27
Leeds NorthWest_5 - Anosh Malik - HTML/CSS - Week 1

AnoshMalik:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -19,7 +19,7 @@ body {
 
 .links{
 display: flex;
-flex-direction: row-reverse;
+flex-direction: row;
 float: right;
 text-decoration: none;
 
@@ -81,6 +81,8 @@ a{
     background-size: cover ;
     background-repeat: no-repeat;
     color: #fff;
+    
+
     height: 735px;
     text-align: center;
     font-family: 'Roboto', sans-serif;
@@ -90,7 +92,7 @@ a{
     justify-content: center;
    
 }
-
+/*something random*/
 .mid{
 
     font-weight: 75px;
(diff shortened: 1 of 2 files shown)


css/store.css
```diff
@@ -1,47 +1,146 @@
...
+.formh1 {
+  color: rgb(252, 109, 57);
+  font-weight: 400;
+  margin-top: 100px;
+}
 
+.mainImageDiv {
+  grid-column: 4/5;
+  grid-row: 1/6;
+  background-image: url("/level-2/storeImage.jpg");
```
levilu800b: I think it would be better you add the image in the html rather than the css.

levilu800b (changes requested):
Please look into that.

css/style.css
```diff
@@ -90,7 +92,7 @@ a{
     justify-content: center;
    
 }
-
+/*something random*/
```
Meronggk, line 74: what's the use of adding something random

Meronggk (changes requested):
what have the random thing added to your code

css/store.css
AnoshMalik: Hmm, that approach was commented over in the html side cuz it wasn't scaling properly

css/style.css
AnoshMalik: Think it was just some random test, I forgot now lol

PR #454, opened 2022-07-28
NW5-Manchester -Ahmed Mohamed-HTML&CSS-Module-project

AhmedMohamed809:

# Your Details

- Your Name:Ahmed Mohamed
- Your City:Manchester
- Your Slack Name:Ahmed Mohamed

# Homework Details

- Module:HTML&CSS-Module-project
- Week: 1

# Notes

- What did you find easy?
- use html and css
- What did you find hard?
use css to move somthing  
 i understand all of the lesson 
- Any other notes?

### DIFF
--- a/index.html
+++ b/index.html
@@ -10,7 +10,74 @@
     <link rel="shortcut icon" type="image/x-icon"  href="favicon.ico">
 </head>
 <body>
+    <main>
+        <header>
+          <div class="main1">
 
+            <img src="img/karma-logo.svg" class="img_logo">
+               <nav calss="menu" >
+                  <a href="#" class="home">Meet Karama</a>
+                  <a href="#" class="home">How it Works</a>
+                  <a  class="active" href="#"  class="home">Store</a>
+                  <a href="#"  class="home">Blog</a>
+                  <a href="#"  class="home">Help</a>
+                  <a href="#"  class="home" >Login</a>
+               </nav>
+        </div>
+</header>
+</main>
+<div class="body1">
+    <h2>
+        Introducing Karma
+    </h2>
+    <h4>Bring Wifi with you, everywhere you go.</h4>
+    <button class="button1" >Learn More</button>
+</div>
+<br><br><br><br><br><br>
+<div class="body2">
+
+    <div class="text2">
+     <h1>Everyone needs a little Karama</h1>
+    </div>
+    
+    </div>
+    <div class="body3">
+    
+        <div class="img3">
+            <img src="img/icon-devices.svg" class="devic">
+            <h4>Internet for all device</h4>
+        </div>
+
+        <div>
+            <img src="img/icon-coffee.svg" class="devic">
+            <h4>Boost your productivity</h4>
+        </div>
+        <div>
+            <img src="img/icon-refill.svg" class="devic">
+            <h4>Pay as You Go </h4>
+        </div>
+    
+
+</div>
+<hr>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -1,19 +1,106 @@
...
+    </div>
+    <div class="contener">
+
+        <img class="img_2" src="level-2/homepage-feature.png">
 
+
+        <div class="text2">
+
+            <h1 class="contentquote"><span class="quote">&ldquo;</span>Wherever I am,I just don't <br>
+                worry about my connection<br> anymore!<span class="quote">&rdquo;</span></h1>
```
AltomHussain, line 70: You should not use <br> tag in within your paragraph, because the br tag makes new line, and you don't need a new line before the paragraph ends, instead make your container flexible so the text can resize and adjust accordingly.

store.html
```diff
@@ -0,0 +1,113 @@
...
+                    <a href="#" class="home">Login</a </ul>
+            </nav>
+        </div>
+        <div class="form1">
+
+            <h1>Order your Karma wifi device today!</h1>
+            <form>
+                <div class="f">
+                    <label for="username1">First name *</label>
+                    <label for="username2">Last name *</label>
```
AltomHussain, line 33: You should not seperate the label form their input, instead the common use of the label is to have it next to its input like this:
 ``` 
 <div>
 <label for="username1">First name *</label>
 <input id="username1" type="text" name="username" />
 </div>

store.html
```diff
@@ -0,0 +1,113 @@
...
+                <label for="address" class="text_ad">Address 2 *</label>
+                <input id="address" type="text" name="address2" />
+
+                <div class="form_c_p">
+                    <label for="city" class="city5">City *</label>
+                    <label for="postcode" class="post5">Postcode *</label>
+                </div>
+
+                <div class="input_c_p">
+                    <select id="city" name="city">
```
AltomHussain, line 54: I think you need a label for your select tag, This text should be the lable `Select your city...` the reason I say this is becaseu if your user don't not select any of your options , the option that says `Select your city...`  could be selected as a default and send to the server which you don't want it to happen. Instead make the first option a blank like this `<option value=""></option>`

store.html
```diff
@@ -0,0 +1,113 @@
...
+                        <option value="Select" selected>Select your city...</option>
+                        <option value="Kartoum">Kartoum</option>
+                        <option value="Kartoum">London</option>
+                    </select>
+
+                    <input id="postcode" type="text" name="postcode" />
+                </div>
+                <br>
+                <label for="color" class="color">Select a colour</label>
```
AltomHussain, line 63: This label should be inside the div

AltomHussain (commented):
If you have time we could review your code and make it better especially the form. 
Please let me know when you are free so we can have a call about this. 

Thanks


Altom

index.html
```diff
@@ -1,19 +1,106 @@
...
+                    <a href="#" class="home">Blog</a>
+                    <a href="#" class="home">Help</a>
+                    <a href="#" class="home">Login</a>
+                </nav>
+            </div>
+        </header>
+    </main>
+    <div class="body1">
+        <p class="p1">
+            <strong> Introducing Karma</strong>
```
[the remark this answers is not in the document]
AhmedMohamed809: really i can't explain  you how your feedback is important to me and effect  my knowledge to be batter .
Thanks Altom again to your feedback

PR #468, opened 2022-07-29
CITY CLASS_ NO -NW5 FIRST KHESIWE_NAME LAST_DUBE  NAME HTML &CSS- MODULE - WEEK_NO 2

KhesiweDube:
The title for your pull request should be made in this format

- Your Name: Khesiwe Dube
- Your City: NW5 Manchester
- Your Slack Name: Khesiwe Dube
- Module: 2
- Week: 2

- What did you find easy? creating nav tag and CSS

- What did you find hard? some css sizing

- What do you still not understand? some CSSand some tags

- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -4,8 +4,22 @@
 body {
     font-family: 'Roboto', sans-serif;
     -webkit-font-smoothing: antialiased;
+    text-align: center;
+    color: black;
 }
-
+header {
+    display: flex;
+    justify-content: space-evenly;
+}
+.join {
+    display: flex;
+    gap: 15px;
+    
+}
+.join img{
+    width: 20px;
+    height: 20px;
+}
 /**
  * Add your custom styles below
  * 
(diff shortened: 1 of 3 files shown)


Ara225:
Hi Khesiwe!
Overall, this looks like a really solid project some really good work here! The look and feel of the site is mostly spot on, especailly on the store page, and you've really tried to use sematitic html throughout.

Did notice a couple genral things as well as a few code specfic things (which I'll comment seprately below)
-  there aren't really any comments in the CSS or HTML- this is a bit of a pain as the CSS has to be quite complex (can't be helped CSS skews that way anyway) so it's a bit hard to get a sense of what it does
- The main page is a bit off: Spacing of the three icons in the second second isn't quite right and the three social media icons are arranged vertically

css/style.css
```diff
@@ -45,43 +49,129 @@ ul li a {
...
+    gap: 20px;
+
+}
+.image-icon img{
+height: 170px;
+
+
 }
 h1 {
     margin-top: 50px;
```
Ara225, line 112: It's best to use rem or em units when possible :)

css/style.css
```diff
@@ -45,43 +49,129 @@ ul li a {
...
 }
+.grid-2 {
+    display: grid;
+    grid-template-columns: 48% 48%;
+    justify-content: space-between;
+}
+.store {
+    color: #DE6E49;
+}
+input[type="text"], input[type="number"], select { display: block; padding: 7px; border: 2px solid #efefef;
```
Ara225, line 125: This is a clever CSS selector! Good work

PR #493, opened 2022-10-25
London9 Group1 - Mana Ghabraei - HTML/CSS - Week 1

ghost:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/css/store.css
+++ b/css/store.css
@@ -1,0 +1,238 @@
+/* store page */
+
+/* main */
+.storemain {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+}
+
+/* section one */
+.formsection {
+  width: 50%;
+  display: flex;
+  flex-direction: column;
+  padding: 2rem;
+  gap: 30px;
+}
+/* form */
+form {
+  display: flex;
+  flex-direction: column;
+  gap: 20px;
+}
+
+/* form title */
+.formsection h1 {
+  color: var(--orange);
+}
+
+/* form classes */
+label {
+  font-size: 12px;
+  font-weight: 400;
+  color: var(--grey-dark);
+}
+
+.fieldset {
+  display: flex;
+  flex-direction: row;
+  align-items: stretch;
+  flex-wrap: wrap;
+  border: none;
+  gap: 30px;
+}
+
+.flex {
+  display: flex;
+  flex-direction: column;
+  flex-wrap: wrap;
+  gap: 5px;
+}
+
+/* input style */
+input[type="text"] {
+  height: 30px;
+  padding: 2px;
+}
+
+input:focus {
+  outline-color: var(--orange) !important;
+}
+
+input[type="checkbox"] {
+  color: var(--orange);
+  background-color: var(--orange);
+}
+
+#postcode {
+  width: 100px;
+  text-transform: uppercase;
+}
+/* select & option */
+select {
+  font-style: italic;
+  font-size: 12px;
+  font-weight: 400;
+  color: var(--grey-dark);
+  width: 250px;
+  height: 30px;
+}
+
+select:focus {
+  outline-color: var(--orange);
+}
+
+/* radio-box */
+.radio-box {
+  display: flex;
+  flex-direction: column;
+  gap: 10px;
+  margin-top: 10px;
+}
+
+.radio-box legend {
+  font-size: 12px;
+  font-weight: 400;
+  color: var(--grey-dark);
+}
+
+.radio-box ul {
+  display: flex;
+  gap: 20px;
+  list-style: none;
+}
+
+/* changed radio-box color */
+.radio-box input[type="radio"] {
+  appearance: none;
+  width: 15px;
+  height: 15px;
+  border: 1px solid var(--grey-dark);
+  border-radius: 50%;
+  background-clip: content-box;
+  padding: 2px;
+  display: inline-block;
+  margin-right: 0.1rem;
+  vertical-align: -3px;
+}
+
+.radio-box input[type="radio"]:checked {
+  background-color: var(--orange);
+}
+
+/* check box */
+.check-box {
+  gap: 2px;
+  margin-top: 20px;
+}
+
+.check-box a {
+  color: var(--orange);
+}
+
+/* change checkbox mark */
+
+.check-box input[type="checkbox"] {
+  appearance: none;
+  width: 15px;
+  height: 15px;
+  border: 1px solid var(--grey-dark);
+  border-radius: 20%;
+  background-clip: content-box;
+  padding: 1px;
+  background-color: var(--white);
+  display: inline-block;
+  margin-right: 0.1rem;
+  vertical-align: -3px;
+}
+
+.check-box input[type="checkbox"]:checked {
+  background-image: url(/level-2/checkmark.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  background-size: cover;
+  background-color: transparent;
+}
+
+/* button */
+button {
+  background-color: var(--orange);
+  color: var(--white);
+  border: 1px solid var(--orange);
+  border-radius: 5px;
+  padding: 10px 25px;
+  transition: all 0.3s ease-in-out;
+}
+
+button:hover {
+  color: var(--orange);
+  background-color: var(--white);
+  border: 1px solid var(--orange);
+}
+/* section two */
+
+.img-section {
+  background-image: url("/level-2/store-image_by-andrew-neel-unsplash.jpg");
+  background-position: center;
+  background-repeat: no-repeat;
+  background-size: cover;
+  width: 50%;
+  margin: 0;
+}
+
+/* tablet and mobile size */
+@media screen and (max-width: 890px) {
+  /* fix textbox names */
+  .fieldset {
+    display: flex;
+    flex-direction: column;
+    align-items: stretch;
+    flex-wrap: wrap;
+  }
+
+  /* fix positions */
+  .store-main {
+    display: flex;
+    flex-direction: column;
+    flex-wrap: wrap;
+    align-items: center;
+  }
+
+  select {
+    width: 320px;
+  }
+
+  .form-section {
+    width: 100%;
+    display: flex;
+    flex-direction: column;
+    padding: 2rem;
+    gap: 30px;
+  }
+
+  .img-section {
+    background-image: url("/level-2/store-image_by-andrew-neel-unsplash.jpg");
+    background-position: center;
+    background-repeat: no-repeat;
+    background-size: cover;
+    height: 100vh;
+    width: 100%;
+    margin: 0;
+    padding: 2rem;
+  }
+
+  .radio-box ul {
+    display: flex;
+    flex-direction: column;
+    gap: 10px;
+    list-style: none;
+  }
+}
+
+/* mobile size */
+@media screen and (max-width: 489px) {
+  .form-section h1 {
+    text-align: center;
+  }
+}
--- a/css/style.css
+++ b/css/style.css
@@ -57,6 +57,9 @@ header {
 .navlinks {
   color: var(--grey);
   text-decoration: none;
+  display: inline-block;
+  text-align: center;
+  position: relative;
 }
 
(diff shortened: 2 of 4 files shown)


css/style.css
```diff
@@ -1,9 +1,20 @@
-
- /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
+/* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
+:root {
+  --grey-dark: #363434;
```
jonnywyatt, line 3: This is a really good use of CSS variables! Did you learn about them in the lesson?

css/style.css
```diff
@@ -15,5 +26,177 @@ body {
  * 
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
+/* header */
+header {
+  display: flex;
```
jonnywyatt, line 31: Good use of flexbox!

css/style.css
```diff
@@ -15,5 +26,177 @@ body {
...
+  justify-content: space-between;
+  align-items: center;
+  padding: 2rem 5rem;
+}
+.logo {
+  display: block;
+  min-width: 2rem;
```
jonnywyatt, line 38: If you give the `img` inside it, a width and height in the HTML, you won't need `min-width` and `min-height`

css/style.css
```diff
@@ -15,5 +26,177 @@ body {
...
+.navlist {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  flex-wrap: wrap;
+  list-style: none;
+}
+
+.navitems {
+  display: flex;
```
jonnywyatt, line 53: No need to set `display: flex` here as that's for use on the flex 'container' (ie .navlist) not its children.

css/style.css
```diff
@@ -15,5 +26,177 @@ body {
...
+  padding: 0 1.5rem;
+}
+
+.navlinks {
+  color: var(--grey);
+  text-decoration: none;
+}
+
+.navlinks:hover {
```
jonnywyatt, line 62: Great 👍

index.html
```diff
@@ -11,9 +11,58 @@
...
+            <ul class="navlist">
+                <li class="navitems"><a class="navlinks" href="#">Meet Karma</a> </li>
+                <li class="navitems"><a class="navlinks" href="#">Store</a> </li>
+                 <li class="navitems"><a class="navlinks" href="#">Blog</a> </li>
+                <li class="navitems"><a class="navlinks" href="#">Help</a> </li>
+                 <li class="navitems"><a class="navlinks" href="#">Login</a> </li>    
+            </ul>
+        </nav>
+    </header> 
+    <p>
```
jonnywyatt, line 26: This should be a `main` tag as it wraps the main page content

index.html
```diff
@@ -11,9 +11,58 @@
...
+        <section class="hero">
```
jonnywyatt, line 27: <section /> is a good choice for these containers (rather than article, because that's for content that could be shared / syndicated with other sites).

index.html
```diff
@@ -11,9 +11,58 @@
...
+            <h2>Introducing Karma</h2>
```
jonnywyatt, line 28: The page doesn't have an `h1`. The text that best describes what the page contains is best set as an `h1`, maybe this heading?

index.html
```diff
@@ -11,9 +11,58 @@
...
+            <h3>Bring WiFi with you, everywhere you go.</h3>
+            <button><a href="#">Learn More</a> </button>
```
jonnywyatt, line 30: A link can't be wrapped in a button. This is best as just a <a />, then style that as a button.

index.html
```diff
@@ -11,9 +11,58 @@
...
+            Everyone needs a little Karma.
+        </h4>
+        <article>
+            <div class="context">
+            <img src="/img/icon-devices.svg" alt="device  icon">
+            <h5>Internet for all devices</h5>
+         </div>
+        <div class="context">
+            <img src="/img/icon-coffee.svg" alt="coffee   icon">
+            <h5>Boost your productivity</h5>
```
jonnywyatt, line 44: These are fine as <div />, they're not really subheadings as there's not content beneath them.

index.html
```diff
@@ -11,9 +11,58 @@
...
+         </div>
+         <div class="context">
+         <img src="/img/icon-refill.svg" alt="refill      icon">
+         <h5>Pay as You Go</h5>
+        </div>
+        </article>
+    </section>
+    </p>
+    <footer>
```
jonnywyatt, line 53: Good use of a semantic tag!

index.html
```diff
@@ -11,9 +11,58 @@
...
+        <h2>Join us on</h2>
+        <ul>
+            <li><a href="#"><img src="/img/twitter-icon.svg" alt="twitter"> </a> </li>
+            <li><a href="#"><img src="/img/facebook-icon.svg" alt="facebook"> </a> </li>
+            <li><a href="#"><img src="/img/instagram-icon.svg" alt="instagram"> </a> </li>
+        </ul>
+        <h5>
```
jonnywyatt, line 60: This is fine as <div />, it's not really a subheading as there's not content beneath it

css/style.css
```diff
@@ -15,5 +26,177 @@ body {
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  list-style: none;
+}
+footer ul li a {
+  display: block;
+  border: 1px solid var(--grey);
+  padding: 1rem;
+  border-radius: 50%;
```
jonnywyatt, line 186: Great! These look close to the design

jonnywyatt (commented):
Great work! Close to the design, and good use of flexbox and semantic markup.

css/style.css
ManaGh: Not all of that I reading an extera book

css/style.css
ManaGh: thank you!

store.html
```diff
@@ -0,0 +1,153 @@
...
+              <label for="addresstwo">Address 2</label>
+              <input type="text" id="addresstwo" />
+            </div>
+          </div>
+
+          <div class="fieldset">
+            <div class="flex">
+              <label for="city">City *</label>
+              <select name="city" id="city" required>
+                <option value="" disabled selected>Select your city...</option>
```
jonnywyatt, line 82: This option shouldn't be disabled, so the user can click on it

store.html
```diff
@@ -0,0 +1,153 @@
...
+
+            <div class="flex">
+              <label for="postcode">Postcode *</label>
+              <input type="text" id="postcode" required />
+            </div>
+          </div>
+
+          <div class="radiobox">
+            <legend>Select a color</legend>
+            <ul>
```
jonnywyatt, line 97: You could add a class to this `ul` so that you can remove the default bullet styling and left padding / margin

store.html
```diff
@@ -0,0 +1,153 @@
...
+              </select>
+            </div>
+
+            <div class="flex">
+              <label for="postcode">Postcode *</label>
+              <input type="text" id="postcode" required />
+            </div>
+          </div>
+
+          <div class="radiobox">
```
jonnywyatt, line 95: This should be a fieldset tag instead of a div, because it's a group of related form controls (radio buttons)

store.html
```diff
@@ -0,0 +1,153 @@
...
+            <button>Place my order</button>
+          </div>
+        </form>
+      </section>
+
+      <section class="imgsection"></section>
+    </main>
+
+    <footer>
+      <h3>Join us on</h3>
```
jonnywyatt, line 138: This should probably be a `h2`. It would be an h3 if it was nested under an existing `h2` section

index.html
```diff
@@ -15,14 +15,21 @@
         <a class="logo" href="index.html"><img src="/img/karma-logo.svg" alt="karma logo"></a>
         <nav>
             <ul class="navlist">
-                <li class="navitems"><a class="navlinks" href="#">Meet Karma</a> </li>
-                <li class="navitems"><a class="navlinks" href="#">Store</a> </li>
+                <input type="checkbox" id="check" />
```
jonnywyatt, line 18: No need for a checkbox here

css/store.css
```diff
@@ -0,0 +1,238 @@
+/* store page */
```
jonnywyatt, line 1: This is good that you've separated out the styles for this page - they'll be easier to manage

css/store.css
```diff
@@ -0,0 +1,238 @@
...
+/* section one */
+.formsection {
+  width: 50%;
+  display: flex;
+  flex-direction: column;
+  padding: 2rem;
+  gap: 30px;
+}
+/* form */
+form {
```
jonnywyatt, line 19: Better to style classes rather than elements, so that you don't get unwanted styles applied to all of the elements

css/store.css
```diff
@@ -0,0 +1,238 @@
...
+  display: flex;
+  flex-direction: column;
+  gap: 20px;
+}
+
+/* form title */
+.formsection h1 {
```
jonnywyatt, line 26: It's better to add a class to the `h1` and target that directly in your CSS. That way, if the `h1` is later moved outside the container, the styling will still be applied

css/store.css
```diff
@@ -0,0 +1,238 @@
...
+  color: var(--orange);
+}
+
+/* form classes */
+label {
+  font-size: 12px;
+  font-weight: 400;
```
jonnywyatt, line 33: No need to set this, as the font-weight is already regular by default, and this also sets it to regular

store.html
```diff
@@ -0,0 +1,153 @@
...
+          /></label>
+        </ul>
+      </nav>
+    </header>
+    <main class="storemain">
+      <section class="formsection">
+        <h1>Order your Karma wifi device today!</h1>
+
+        <form action="#">
+          <div class="fieldset">
```
jonnywyatt, line 53: Calling this class fieldset is a bit confusing as the div isn't a fieldset. The [fieldset tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset) is used to group several checkboxes or radio buttons together.

css/store.css
```diff
@@ -0,0 +1,238 @@
...
+  margin-top: 20px;
+}
+
+.check-box a {
+  color: var(--orange);
+}
+
+/* change checkbox mark */
+
+.check-box input[type="checkbox"] {
```
jonnywyatt, line 136: This styles aren't being picked up by the page because the class in the page is `checkbox` not `check-box`

css/store.css
```diff
@@ -0,0 +1,238 @@
...
+  background-image: url("/level-2/store-image_by-andrew-neel-unsplash.jpg");
+  background-position: center;
+  background-repeat: no-repeat;
+  background-size: cover;
+  width: 50%;
+  margin: 0;
+}
+
+/* tablet and mobile size */
+@media screen and (max-width: 890px) {
```
jonnywyatt, line 185: Well done for using a media query!

You don't need to repeat all the CSS inside the media query, only the parts that will change

css/store.css
```diff
@@ -0,0 +1,238 @@
...
+  /* fix textbox names */
+  .fieldset {
+    display: flex;
+    flex-direction: column;
+    align-items: stretch;
+    flex-wrap: wrap;
```
jonnywyatt, line 191: No need to use flex-wrap in any of these selectors, unless you want the boxes to wrap? (they don't in the design)

css/store.css
```diff
@@ -0,0 +1,238 @@
...
+
+  .form-section {
+    width: 100%;
+    display: flex;
+    flex-direction: column;
+    padding: 2rem;
+    gap: 30px;
+  }
+
+  .img-section {
```
jonnywyatt, line 214: Is this being used anywhere on the store page?

jonnywyatt:
Some good work here! Well done for adding a media query!
I can't see a commit for the new section on index.html?
Also, all the images are broken on the store page - can you run locally with VS Code live server, to see? The folder names you're using in your `img` src attributes are wrong I think.

PR #498, opened 2022-10-26
London class 9 -Mickey Haile -HTML/CSS -week 1

mickeyhaile2:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -15,241 +15,81 @@ body {
  * 
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
- main > section:nth-child(1) {
-    display: flex;
-    flex-direction: column;
-    justify-content: center;
-    align-items: center;
-    gap: 1rem;
-    padding: 8rem 0;
-    background-image: url("../img/first-background.jpg");
-    background-size: cover;
-    background-repeat: no-repeat;
-    background-position: center;
-    color: white;
-  }
-  
-  main > section:nth-child(1) > h1 {
-    margin: 0; 
-    /* this shouldn't be neccessary when i globally set margin to 0 at the top.. */
-    /* is normalize messing with this? but styles.css is loaded AFTER normalize? */
-    font-size: 2.8rem;
-  }
-  
-  main > section:nth-child(1) > h3 {
-    font-size: 1.6rem;
-  }
-  
-  main > section:nth-child(1) > a {
-    margin-top: 2rem;
-    padding: 1rem 2rem;
-    border-radius: 0.3rem;
-    background-color: hsl(15, 87%, 55%);
-    color: inherit;
-    text-decoration: none;
-  }
-  
-  /* SECTION TWO */
-  main > section:nth-child(2) {
-    display: flex;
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -15,5 +15,113 @@ body {
...
+.content,
+.footer {
+  padding: 0 1.5rem;
+  max-width: 1200px;
+  margin: 0 auto;
+  display: flex;
+    justify-content: space-between;
+    align-items: center;
+    padding: 2rem 5rem;
+}
```
JDysiewicz, line 39: nit: rather than applying these stylings to all 3 classes, you could instead apply them to a single class (e.g. `.text-block`), and then apply that class to each of your header/content/footer (e.g. `<div class="header text-block">`) - that way you're not coupling the styling of the header/footer/content to each other

css/style.css
```diff
@@ -15,5 +15,113 @@ body {
...
+  text-align: center;
+  flex-wrap: wrap;
+  
+}
 
+/*footer*/
+.footer1 {
+  
+  
+}
```
JDysiewicz, line 106: no need to include unnecessary code

css/style.css
```diff
@@ -15,5 +15,113 @@ body {
...
+  font-size: 0.5rem;
+  display: flex;
+}
+button {
+    background-color: #c05326;
+
+}
+.hi .hi2 .hi3 {
+    padding: 0 15px;
+}
```
JDysiewicz, line 127: try to have logical names for classes/variables; the classes/variable names should indicate what they represent so as not to rely on future developers having all the context to understand them

css/style.css
```diff
@@ -15,5 +15,113 @@ body {
...
+  
+  
+}
 
+.footer2 {
+  border-top: 1px solid var(--grey-light);
+  padding-top: 2rem;
+  padding-left: 550px;
+  padding-bottom: 2rem;
+
```
JDysiewicz, line 112: nice use of padding to create negative space :)

JDysiewicz (approved):
looks good to me, well done

css/store.css
```diff
@@ -0,0 +1,85 @@
+body {
+    font-family: 'Roboto', sans-serif;
+    -webkit-font-smoothing: antialiased;
+  font-family: "Roboto", sans-serif;
+  -webkit-font-smoothing: antialiased;
+  text-align: center;
+  background-image: url(C:\Users\User\Documents\GitHub\HTML-CSS-Module-Project\img\first-background.jpg);
```
JDysiewicz, line 7: this path won't resolve correctly; will need to be something like `/img/first-background.jpg`

JDysiewicz:
As a general note, I'd avoid naming your file with spaces in them (e.g. `store page.html`). It can make pathing between things pretty complicated and will likely cause issues in the future - if you want to put spaces in your file names, I'd advise using `-` or `_` instead (e.g. `store-page.html` or `store_page.html`)

PR #502, opened 2022-10-26
London 9 - Andriana Saffo - HTML/CSS - Week 1

AndrianaOS:

# Your Details

- Your Name: Andriana Saffo
- Your City: London
- Your Slack Name: Andriana

# Homework Details

- Module: HTML/CSS
- Week: 1

# Notes

- What did you find easy?
- Creating the general information and the simple styling

- What did you find hard?
- Getting the background image to render on the page

- N/A

- Any other notes?
- N/A


css/style.css
```diff
@@ -16,4 +15,149 @@ body {
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
 
+/*HEADER STYLES*/
+
+#karma-logo {
+  width: 25px;
+  margin-left: 50px;
+  margin-top: 50px;
+  margin-bottom: 20px;
```
bazmurphy, line 24: if you want to make this 1 line instead of 3 you can use the shorthand:
margin: 50px 0 20px 50px;
[top right bottom left]

store.css
```diff
@@ -0,0 +1,199 @@
...
+  margin-top: 50px;
+  font-size: 20px;
+  width: fit-content;
+}
+
+/*STYLING  FOR FOOTER*/
+
+footer {
+  text-align: center;
+  margin: 0 70px 0 70px;
```
bazmurphy, line 183: this can be shortened:
margin: 0 70px;
[top/bottom left/right]

store.html
```diff
@@ -0,0 +1,164 @@
...
+  <body>
+    <header>
+      <img src="img/karma-logo.svg" alt="karma logo" id="karma-logo" />
+      <nav id="nav-bar">
+        <a href="index.html" id="karma-meet">Meet Karma</a>
+        <a href="works">How it Works</a>
+        <a href="store">Store</a>
+        <a href="blog">Blog</a>
+        <a href="help">Help</a>
+        <a href="login">Login</a>
```
bazmurphy, line 25: semantically: this navigation should be an unordered list of list items that contain anchors
this is helpful for accessibility
```
<ul>
  <li>
    <a>
```

store.html
```diff
@@ -0,0 +1,164 @@
...
+        </div>
+      </section>
+    </main>
+    <footer>
+      <p>Join us on</p>
+      <div>
+        <img src="img/twitter-icon.svg" alt="twitter logo" class="smicon" />
+        <img src="img/facebook-icon.svg" alt="facebook logo" class="smicon" />
+        <img src="img/instagram-icon.svg" alt="instagram logo" class="smicon" />
+      </div>
```
bazmurphy, line 160: semantically: this div should instead be an unordered list of list items with anchor tags that wrap images (in that order)
(remember a div has no semantic meaning)
so for accessibility it is understood that it is a list of external social media links

```
<ul>
  <li>
    <a>
      <img>
```

store.css
```diff
@@ -0,0 +1,199 @@
...
+section {
+  display: flex;
+  justify-content: space-between;
+  border-top: 1px solid #ebebeb;
+  width: 100%;
+  gap: 50px;
+}
+
+.title-form {
+  flex-direction: column;
```
bazmurphy, line 72: this is missing: 
display: flex
because flex-direction: column on its own will not do anything to a container that has not been defined as a flexbox

store.html
```diff
@@ -0,0 +1,164 @@
...
+              </div>
+              <label for="terms"
+                >By placing your order you agree to Karma's<a
+                  href="termsconditions"
+                  >Terms and Conditions</a
+                >*</label
+              >
+            </div>
+
+            <button>Place my order</button>
```
bazmurphy, line 143: Interesting to know:

When inside a `<form>` :
`<input type="submit">Place my order</input>`
Will create a button thats function is to submit the form

And also when inside a `<form>`
a single `<button>` with no defined type will default to "submit"

Whereas in this case:
```
<button type="submit">Submit</button>
<button type="reset">Reset</button>
```
If you have multiple buttons, it is important to explicitly name their type.

So yours would be:
`<button type="submit">Place my order</button>`

I know this does not apply in this case (because there is only a single button) but it is good to know for later on.

css/style.css
```diff
@@ -16,4 +15,149 @@ body {
...
+h2 {
+  text-align: center;
+  color: white;
+  font-weight: lighter;
+  font-size: 40px;
+  margin-top: 0;
+}
+
+button {
+  align-content: center;
```
bazmurphy, line 63: `align-content` is a property used to adjust flex/grid cells content horizontally/vertically
in order for this to work the element must also be declared as `display: flex` or `display: grid`
but the `button` is not currently a flex/grid container
maybe i am wrong, but i am assuming you are trying to center the button text with this?

css/style.css
```diff
@@ -16,4 +15,149 @@ body {
...
+
+#karma-logo {
+  width: 25px;
+  margin-left: 50px;
+  margin-top: 50px;
+  margin-bottom: 20px;
+}
+
+nav {
+  float: right;
```
bazmurphy, line 28: i would recommend trying to make your `<header>` a flex box
and the `` (logo) and the `<nav>` will be the flex children
then separate the flex children using `justify-content: space-between` to push them to either side of the page, the image to the left and the nav to the right
you can then apply appropriate padding to flex parent
using` float` can be quite tricky and cause a lot of headaches.

bazmurphy (commented):
Here is some feedback and suggestions 😎

PR #504, opened 2022-10-27
London Class 9 - Vitalina Kuzmenko - HTML/CSS - Week 1

VitalinaKuzmenko:
https://html-css-module-project.netlify.app/


index.html
```diff
@@ -1,19 +1,120 @@
...
     <link href='https://fonts.googleapis.com/css?family=Roboto:400,500,300' rel='stylesheet' type='text/css'>
     <link rel="stylesheet" href="css/normalize.css">
     <link rel="stylesheet" href="css/style.css">
-    <link rel="shortcut icon" type="image/x-icon"  href="favicon.ico">
+    <link rel="shortcut icon" type="image/x-icon" href="favicon.ico">
 </head>
+
 <body>
+    <header>
+        <nav class="navbar navbar-expand-lg navbar-light bg-light fixed-top menu">
```
Tony-devops, line 18: good utilisation of bootstrap for responsive and hamburger menu

index.html
```diff
@@ -1,19 +1,120 @@
...
+        </div>
+    </footer>
+
+    <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.2.0/dist/js/bootstrap.bundle.min.js"
+        integrity="sha384-A3rJD856KowSb7dwlZdYEkO39Gagi7vIsF0jrRAoQmDKKtQBHUuLZ9AsSv4jD4Xa"
+        crossorigin="anonymous"></script>
+    <script src="https://kit.fontawesome.com/e8c9266318.js" crossorigin="anonymous"></script>
 </body>
-</html>
+
```
Tony-devops, line 119: good use of semantic tags.

calarasiu:
The website looks very nice, although try to achieve all of that using only css, since that's the goal of the module. It's for sure a point in your favor knowing bootstrap, but use only css for the coursework. 
I only have a few suggestions:

- the form is better to make it fit in a pageview - so the user knows immediately how many fields he has to complete

- when the fields are required is good practice to give the feedback to the user on which fields needs to be completed before he can submit 
Other than that amazing job 👏

index.html
VitalinaKuzmenko: Thank you, Tony!

index.html
VitalinaKuzmenko: Thank you, Tony!

PR #516, opened 2022-10-28
LONDON CLASS_9-SAGHAR_HOSSEINMARDI-HTML/CSS-WEEK_2

SagharHosseinmardi:

# Your Details

- Your Name: Saghar Hosseinmardi
- Your City: London
- Your Slack Name:Saghar Hosseinmardi

# Homework Details

- Module: HTML/CSS
- Week: 2

# Notes

- What did you find easy?
Nothing was easy in this course work for me 
- What did you find hard?
Flex-box
flex-box and change the spaces size 
- Any other notes?


css/style.css
```diff
@@ -15,5 +14,195 @@ body {
  * 
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
+h1 {
+  margin: auto;
+  margin-top: auto;
+  font-size: 40px;
```
jonnywyatt, line 20: As we discussed last week, using pixels for font sizes is fine for the moment. rem's are better for accessibility as the font size will change when the user updates their text size in their browser settings. 1 rem equals 16px in all main browsers.

css/style.css
```diff
@@ -15,5 +14,195 @@ body {
...
+}
+h2 {
+  margin: auto;
+  margin-top: auto;
```
jonnywyatt, line 24: the `margin` property above sets that margin on all 4 sides of the box, so no need to set again with `margin-top`

jonnywyatt:
There are some really good things in this pull request, well done!
Firstly, the page in the browser looks pretty close to the [original design](https://github.com/CodeYourFuture/HTML-CSS-Module-Project/blob/master/homepage-desktop.png).
Also, you've used flexbox very well. This is a really useful technique for simple layouts so it's great that you've grasped it so quickly.

index.html
```diff
@@ -1,19 +1,106 @@
...
+    <!--header starts here-->
+    <header class="header">
+      <img src="img/karma-logo.svg" alt="karma-logo" class="navbar-logo" />
 
-    <!-- Add your HTML markup here -->
-    <!-- Remember: Use semantic HTML tags like <header>, <main>, <nav>, <footer>, <section> etc -->
-    <!-- All the images you need are in the 'img' folder -->
+      <nav>
+        <ul class="navbar-list">
+          <li id="navbar-button"><a href="#">Meet Karma</a></li>
```
jonnywyatt, line 27: Every`id` attribute in an HTML page has to be unique; this will work fine as a `class` though.

css/style.css
```diff
@@ -15,5 +14,195 @@ body {
...
+}
+
+#navbar-button {
+  margin-left: 25px;
+  margin-right: 25px;
+}
+
+.header {
+  display: flex;
+  flex-direction: row;
```
jonnywyatt, line 40: `flex-direction: row` is the default so no need to add (tho it doesn't do any harm either!)

css/style.css
```diff
@@ -15,5 +14,195 @@ body {
...
+  align-content: space-between;
+  margin: 15px;
+}
+
+.navbar-logo {
+  display: flex;
+  margin: auto;
+  width: 2%;
```
jonnywyatt, line 48: For a logo with a fixed size, maybe simpler to set a width in pixels

css/style.css
```diff
@@ -15,5 +14,195 @@ body {
...
+  margin-left: 200px;
+}
+
+.navbar-list {
```
jonnywyatt, line 52: Good use of flexbox!

css/style.css
```diff
@@ -15,5 +14,195 @@ body {
...
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  list-style-type: none;
+  font-size: smaller;
+  margin-right: 60px;
+}
 
+a:hover {
```
jonnywyatt, line 61: Great 👍 

css/style.css
```diff
@@ -15,5 +14,195 @@ body {
...
+  color: red;
+}
+
+.background-image {
```
jonnywyatt, line 65: These properties are for CSS background images. You've used a `` element in the page (which is fine 👍 ) so these CSS properties will have no effect on it (it looks fine without them).

index.html
```diff
@@ -1,19 +1,106 @@
...
+          <li id="navbar-button"><a href="#">Blog</a></li>
+          <li id="navbar-button"><a href="#">Help</a></li>
+          <li id="navbar-button"><a href="#">Login</a></li>
+        </ul>
+      </nav>
+    </header>
+    <!--main starts here-->
+    <main>
+      <section class="main-articles">
+        <article class="article-one">
```
jonnywyatt, line 39: `<article />` is more for content that will be shared on other sites; so `<section />` is probably a better fit here.

index.html
```diff
@@ -1,19 +1,106 @@
...
+          <img
+            src="img/first-background.jpg"
+            alt="first-background"
+            class="background-image"
+          />
+          <h1 class="line-one">Introducing Karma</h1>
```
jonnywyatt, line 45: I agree, this is the best choice for the `h1` as it best describes what the page is about.
Remember the `h1` doesn't have to be the text with the biggest font (tho it is here).

index.html
```diff
@@ -1,19 +1,106 @@
...
+          <h2 class="line-two">Bring WiFi with you, everywhere you go.</h2>
+          <button class="button">Learn More</button>
```
jonnywyatt, line 47: Even though this looks like a button, it will probably link to another page, so the best choice of tag would be `<a />`

index.html
```diff
@@ -1,19 +1,106 @@
...
+        </article>
 
-</body>
+        <article class="article-two">
+          <h3 class="little-karma">Everyone Needs a Little Karma.</h3>
+          <ul class="icones-list">
+            <li class="photo-text">
+              <img
+                src="img/icon-devices.svg"
+                alt="devices-icon"
```
jonnywyatt, line 56: Alt text is to be read out eg by a screenreader, so should be a normal sentence that describes the image eg 'Picture of some devices'

index.html
```diff
@@ -1,19 +1,106 @@
...
+              <img src="img/icon-refill.svg" alt="refill-icon" class="refill" />
+              <a href="#">Pay as You Go</a>
+            </li>
+          </ul>
+        </article>
+      </section>
+    </main>
+    <!--footer starts here-->
+    <hr />
+    <footer class="footer">
```
jonnywyatt, line 75: Good use of a semantic tag!

index.html
```diff
@@ -1,19 +1,106 @@
...
+        <li class="icons">
+          <a href="#">
+            <img
+              src="img/instagram-icon.svg"
+              alt="instagram-icon"
+              class="instagram-icon"
+          /></a>
+        </li>
+      </ul>
+      <h6 class="mobility">@ Karma Mobility, Inc.</h6>
```
jonnywyatt, line 103: This doesn't need to be a heading as there's no content after it. `<p />` or `<div />` or `<small />` is fine

css/style.css
```diff
@@ -15,5 +14,195 @@ body {
...
+  font-size: 20px;
+  font-weight: lighter;
+}
+
+.socialmedia-list {
+  list-style-type: none;
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  margin-right: 45px;
```
jonnywyatt, line 189: `<ul />` comes with some default padding and margin on the left side. If you reset those to 0 it will have the same effect as this 👍

css/style.css
```diff
@@ -15,5 +14,195 @@ body {
...
+}
+
+.twitter-icon,
+.facebook-icon,
+.instagram-icon {
+  width: 20px;
+  height: 20px;
+  border-radius: 50%;
```
jonnywyatt, line 197: If you set this on the `<a />` that wraps it instead, you can also set a border on it and it will look like the design.

jonnywyatt:
Good work! 
A tip on commit messages - for each commit message, try to briefly describe what the change was, rather than 'Update index.html'. Otherwise every commit message will look similar and you won't be able to tell the difference when you look at your commit history (you can see it listed on this page)

css/store.css
```diff
@@ -0,0 +1,196 @@
...
+}
+
+.field-4 {
+  display: flex;
+  flex-direction: column;
+  width: 40%;
+  margin-left: 5%;
+}
+
+h2 {
```
jonnywyatt, line 118: Better to target a class rather than every `h2`, so that unwanted styles aren't applied to every `h2`

css/store.css
```diff
@@ -0,0 +1,196 @@
...
+  font-weight: 120;
+  font-size: 1vw;
```
jonnywyatt, line 120: Simpler to use `px` or `rem` for font-size (preferably `rem`, for accessibility)

store.html
```diff
@@ -0,0 +1,3160 @@
...
+                <option value="Youlgreave">Youlgreave</option>
+                <option value="Yoxall">Yoxall</option>
+              </select>
+            </div>
+            <div class="field-4">
+              <label for="postcode">Postcode</label>
+              <input required type="text" name="postcode" />
+            </div>
+          </div>
+          <h2>select a colour</h2>
```
jonnywyatt, line 3085: There should be a `fieldset` wrapping this set of radio buttons. This `h2` should be a `legend` tag instead.

store.html
```diff
@@ -0,0 +1,3160 @@
...
+                name="space-gray"
+              />
+              <label for="space-gray" class="radio-margin">Space Gray</label>
+            </div>
+          </div>
+          <div class="checkbox">
+            <input type="checkbox" name="checkbox" />
+            <label for="checkbox"
+              >By placing your order you agree to Karma's
+              <span url="#" style="color: orange">Term's and Conditions</span
```
jonnywyatt, line 3112: This should be an `a` tag with an `href` attribute instead of `url`

store.html
```diff
@@ -0,0 +1,3160 @@
...
+              <label for="karma-orange" class="radio-margin"
+                >Karma orange</label
+              >
+            </div>
+            <div>
+              <input
+                required
+                type="radio"
+                value="space-gray"
+                name="space-gray"
```
jonnywyatt, line 3103: Every input should have an `id` attribute that has the same value as the `for` attribute on its label

store.html
```diff
@@ -0,0 +1,3160 @@
...
+            <div class="field-2">
+              <label for="address">Address 2</label>
+              <input required type="text" name="address2" />
+            </div>
+          </div>
+          <div class="city">
+            <div class="field-3">
+              <label for="city">City</label>
+              <!-- England-->
+              <select id="city" name="city">
```
jonnywyatt, line 65: At the moment this is the width of its largest option. You could use CSS to set the `select` to be 100% of the width of its parent element.

store.html
```diff
@@ -0,0 +1,3160 @@
...
+              <span url="#" style="color: orange">Term's and Conditions</span
+              >.</label
+            ><br />
+          </div>
+
+          <button class="button-order">Place my order</button>
+        </div>
+        <img
+          src="level-2/store-image_by-andrew-neel-unsplash.jpg"
+          alt="a woman working with laptop"
```
jonnywyatt, line 3121: Good alt text!

jonnywyatt (commented):
Good job!

index.html
```diff
@@ -1,19 +1,125 @@
...
+      <ul class="socialmedia-list">
+        <li class="icons">
+          <a href="#">
+            <img
+              src="img/twitter-icon.svg"
+              alt="twitter-icon"
+              class="twitter-icon"
+          /></a>
+        </li>
+        <li class="icons">
```
KhadijaHaji, line 105: Your formatting throughout looks really good!

css/store.css
```diff
@@ -0,0 +1,196 @@
...
+ * Add your custom styles below
+ * 
+ * Remember: 
+ * - Be organised, use comments and separate your styles into meaningful chunks
+ *    for example: General styles, Navigation styles, Hero styles, Footer etc.
+ * 
+ * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
+ */
+
+#navbar-button {
```
KhadijaHaji, line 18: good to see you're using both id's and classes.

PR #517, opened 2022-10-28
Leila Farsani-Karma-Clone-Week2

leilafarsani:

# Your Details

- Your Name: Leila Farsani
- Your City: London
- Your Slack Name: Leila Farsani

# Homework Details

- Module: HTML-CSS-Module-Project
- Week: 2

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


index.html
```diff
@@ -10,10 +10,54 @@
...
+            <li class="items"><a class="links" href="#"> How it Works</a></li>
+            <li class="items"><a class="links" href="#"> Store</a></li>
+            <li class="items"><a class="links" href="#"> Blog</a></li>
+            <li class="items"><a class="links" href="#"> Help</a></li>
+            <li class="items"><a class="links" href="#"> Login</a></li>  
+        </ul>     
+    </nav>
+ </header> 
+ <main> 
+    <section class="p1">
```
jonnywyatt, line 27: Really good clean markup structure, well done. You've used header, main, footer at the top level. `<section />` is a good choice for the containers below that (rather than article, because that's for content that could be shared / syndicated with other sites).

index.html
```diff
@@ -10,10 +10,54 @@
...
+        <h1>Introducing Karma</h1>
+        <h4>Bring WiFi with you, everywhere you go.</h4>
+        <button><a href="#">Learn More</a></button>
+    </section>
+    <section>
+        <h2>
+            Everyone needs alitte Karma.
+        </h2>
+        <div class="features">
+            <div class="p2">
```
jonnywyatt, line 37: Nearly all your CSS classnames are really clear and easy to understand. Maybe worth changing `p1` and `p2` as well so when you see them in the CSS you'll remember what they're for.

css/style.css
```diff
@@ -1,19 +1,189 @@
...
+}
+.p1 button a {
+    text-decoration: none;
+    color: var(--white);
+}
+section {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    margin-top: 4rem;
```
jonnywyatt, line 110: This is why you also needed a negative -60px margin on `main`, above.
Simpler to not style all sections, use a class selector instead and just target the sections you need to add this margin to, rather than all sections.

css/style.css
```diff
@@ -1,19 +1,189 @@
...
+
+.items:first-child > a{
+color: var(--grey-darkdark);
+}
+.items:first-child > a:hover{
+    color: var(--orange);
+}
+.p1 {
+    background-image: url(/img/first-background.jpg);
+    background-size: cover;
```
jonnywyatt, line 70: Great!

css/style.css
```diff
@@ -1,19 +1,189 @@
...
+    background-repeat: no-repeat;
+    background-position: center;
+    padding: 12rem 0;
+    color: var(--white);
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+    gap: none;
```
jonnywyatt, line 79: Easier to just leave this property out. I don't think a value of none will be understood by the browser anyway.

css/style.css
```diff
@@ -1,19 +1,189 @@
...
-/**
- * Add your custom styles below
- * 
- * Remember: 
- * - Be organised, use comments and separate your styles into meaningful chunks
- *    for example: General styles, Navigation styles, Hero styles, Footer etc.
- * 
- * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
- */
+.links:hover{
```
jonnywyatt, line 58: Great 👍

index.html
```diff
@@ -10,10 +10,54 @@
...
+            <li class="items"><a class="links" href="#"> Blog</a></li>
+            <li class="items"><a class="links" href="#"> Help</a></li>
+            <li class="items"><a class="links" href="#"> Login</a></li>  
+        </ul>     
+    </nav>
+ </header> 
+ <main> 
+    <section class="p1">
+        <h1>Introducing Karma</h1>
+        <h4>Bring WiFi with you, everywhere you go.</h4>
```
jonnywyatt, line 29: With the heading structure in a page, it's not valid to jump several levels like this from `h1` to `h4`, so best as either a `h2` or a `div`

index.html
```diff
@@ -10,10 +10,54 @@
...
+        <button><a href="#">Learn More</a></button>
```
jonnywyatt, line 30: A link can't be wrapped in a button. This is best as just a `<a />`, then style that as a button.

index.html
```diff
@@ -10,10 +10,54 @@
...
+    </section>
+    <section>
+        <h2>
+            Everyone needs alitte Karma.
+        </h2>
+        <div class="features">
+            <div class="p2">
+              <img  src="img/icon-devices.svg" alt="">
+              <h5>Internet for all devices </h5>
```
jonnywyatt, line 39: These are fine as `<div />`, they're not really subheadings as there's not content beneath them.

index.html
```diff
@@ -10,10 +10,54 @@
...
+        <h4>Bring WiFi with you, everywhere you go.</h4>
+        <button><a href="#">Learn More</a></button>
+    </section>
+    <section>
+        <h2>
+            Everyone needs alitte Karma.
+        </h2>
+        <div class="features">
+            <div class="p2">
+              <img  src="img/icon-devices.svg" alt="">
```
jonnywyatt, line 38: I think this could maybe use an alt, but interesting discussion to have

index.html
```diff
@@ -10,10 +10,54 @@
...
+            </div>
+            <div class="p2">
+                <img  src="img/icon-refill.svg" alt="">
+                <h5>Pay as You Go</h5>
+            </div>
+        </div>
+    </section>
+ </main> 
+ <footer>
+   <span><h2>Join us on</h2></span>
```
jonnywyatt, line 53: No need to wrap a block element like `h2` in a `span`, it will have no effect.

jonnywyatt (commented):
Really good work - it looks like the design, and you've used flexbox and semantic HTML effectively!

index.html
```diff
@@ -10,10 +10,54 @@
...
+                <h5>Boost your productivity</h5>
+            </div>
+            <div class="p2">
+                <img  src="img/icon-refill.svg" alt="">
+                <h5>Pay as You Go</h5>
+            </div>
+        </div>
+    </section>
+ </main> 
+ <footer>
```
jonnywyatt, line 52: Good use of a semantic tag!

css/style.css
```diff
@@ -1,189 +1,233 @@
...
-
- .logo{
-    display: block;
-    min-width: 1.5rem;
-    min-height: 1.5
-    rem;
- }
+  display: flex;
+  flex-direction: column;
+  margin-top: -80px;
```
jonnywyatt, line 29: The layout will be simpler to develop if you remove this line, and also remove the [margin-top on the `section` selector](https://github.com/CodeYourFuture/HTML-CSS-Module-Project/pull/517/commits/ddbbba3580693f159c7b26eb2d110ae20542af4d#diff-1fc556f95754ee7e33d91044125c44bb9f750c99be4406756ffb27413adfcaf5R109).
Remember, it's best not to style using element selectors, as it will affect every one of them and then you might have to add more CSS to work against those styles, as you have here.

css/style.css
```diff
@@ -1,189 +1,233 @@
...
+}
+
+.p3-text a:hover {
+  background-color: var(--yellow);
+}
+
+.part3 {
+  display: grid;
+  grid-template-rows: auto;
+  grid-template-columns: 2fr 3fr;
```
jonnywyatt, line 202: great! Good use of grid here

css/style.css
```diff
@@ -1,189 +1,233 @@
...
+  background-color: #ffdfdf;
+}
+
+.p3-img {
+  background-color: #363434;
```
jonnywyatt, line 207: Try to use your colour variables everywhere

css/style.css
```diff
@@ -1,189 +1,233 @@
...
+  background-image: url(/level-2/homepage-feature.png);
+  background-repeat: no-repeat;
+  background-size: cover;
+}
+
+.p3-text {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  gap: 2rem;
```
jonnywyatt, line 218: Great! This spaces out the two elements in the flex container

css/style.css
```diff
@@ -231,3 +231,97 @@ footer h5 {
 .p3-text p {
   font-style: italic;
 }
+
+.main2 {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
```
jonnywyatt, line 238: I don't think this line will help; you don't want the second section to wrap to the next line

css/style.css
```diff
@@ -231,3 +231,97 @@ footer h5 {
...
+}
+
+form {
+  display: flex;
+  flex-direction: column;
+  gap: 20px;
+}
+
+.section2 {
```
jonnywyatt, line 247: This is working fine, but your selector which targets all `section` elements is adding margin-top that you don't need

css/style.css
```diff
@@ -231,3 +231,97 @@ footer h5 {
...
+  gap: 30px;
+}
+
+.section2 h1 {
+  color: var(--orange);
+}
+
+label {
+  font-size: 12px;
+  font-weight: 400;
```
jonnywyatt, line 261: This `font-weight` will have no effect as the font is set to regular weight by default anyway, which is the same as `400`

css/style.css
```diff
@@ -231,3 +231,97 @@ footer h5 {
...
+.fieldset {
+  display: flex;
+  flex-direction: row;
+  align-items: stretch;
+  flex-wrap: wrap;
+  border: none;
+  gap: 30px;
+}
+
+.what {
```
jonnywyatt, line 274: Try to name your classes with a bit more meaning eg `.labelInputWrapper` or something

css/style.css
```diff
@@ -231,3 +231,97 @@ footer h5 {
...
+  display: flex;
+  flex-direction: column;
+  flex-wrap: wrap;
+  gap: 5px;
+}
+
+select {
```
jonnywyatt, line 281: Try to use classes rather than targetting all elements

css/style.css
```diff
@@ -231,3 +231,97 @@ footer h5 {
...
+  font-size: 12px;
+  font-weight: 400;
+  color: var(--grey-dark);
+  width: 170px;
+  height: 23px;
```
jonnywyatt, line 286: No need to set height here as it will just be its natural height

css/style.css
```diff
@@ -231,3 +231,97 @@ footer h5 {
...
+  align-items: stretch;
+  flex-wrap: wrap;
+  border: none;
+  gap: 30px;
+}
+
+.what {
+  display: flex;
+  flex-direction: column;
+  flex-wrap: wrap;
```
jonnywyatt, line 277: No need for flex-wrap

css/style.css
```diff
@@ -231,3 +231,97 @@ footer h5 {
...
+
+.check-box {
+  gap: 2px;
+}
+
+.check-box a {
+  color: var(--orange);
+}
+
+button {
```
jonnywyatt, line 314: Use classes, don't target elements

css/style.css
```diff
@@ -231,3 +231,97 @@ footer h5 {
...
+  font-size: 12px;
+  font-weight: 400;
+  color: var(--grey-dark);
+}
+
+.radio-box ul {
+  display: flex;
+  gap: 10px;
+  list-style: none;
+}
```
jonnywyatt, line 304: `ul` has some left margin and padding by default, you could reset them to zero here to make it look closer to the design.

index.html
```diff
@@ -66,7 +66,140 @@ <h5>Pay as You Go</h5>
...
+    </nav>
+ </header> 
+
+ <main class="main2">
+
+    <section class="section2">
+        <h1>Order your Karma wifi device today!</h1>
+
+        <form action="#">
+            <div class="fieldset">
```
jonnywyatt: Call this class `fieldset` is a bit confusing as the div isn't a fieldset. The [`fieldset` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset) is used to group several checkboxes or radio buttons together.

index.html
```diff
@@ -66,7 +66,140 @@ <h5>Pay as You Go</h5>
...
+                  <label for="postcode">Postcode *</label>
+                  <input 
+                    type="text"
+                    id="postcode"
+                    required
+                  />
+                </div>
+            </div>
+
+            <div class="radio-box">
```
jonnywyatt: This should be a `fieldset` tag instead of a `div`, because it's a group of related form controls (radio buttons)

index.html
```diff
@@ -66,7 +66,140 @@ <h5>Pay as You Go</h5>
...
+            </div>
+
+            <div>
+                <button>Place my order</button>
+            </div>
+        </form>
+    </section>
+
+    <section class="section2-img">
+        <img src="level-2/store-image_by-andrew-neel-unsplash.jpg" alt="">
```
jonnywyatt: Maybe this is a decorative image, like the one in the new section on index.html, so you could use CSS `background-image` to add it

css/style.css
```diff
@@ -231,3 +231,97 @@ footer h5 {
 .p3-text p {
   font-style: italic;
 }
+
+.main2 {
```
jonnywyatt, line 235: In the design, the two sections under the `main` tag share the width 50% each. How can you do that here? Maybe `grid` would be simpler, to do that?

PR #519, opened 2022-10-28
London9_lovelace- Boshra_Mahmoudi_MODULE-WEEK-1

BoshraM:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -38,7 +38,7 @@ header {
 .logo_header {
   width: 2rem;
   height: 2rem;
-  margin-right: 5rem;
+
   padding: 1rem;
 }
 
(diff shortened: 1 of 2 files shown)


VitalinaKuzmenko (commented):
1st page
1)	It is better to put section (class=extended_section) inside main tag. Everything else is really good. The structure of html is very good! Nice job!
2)	I understand why you did the page of size – 1440px. But I think we should do it on whole screen. For example, my screen is wider than 1440 px, that’s why I have white strap from left and right.
3)	I see that you’re using a lot as a size – px. Just an advice – it makes your website not responsive at all. And the body of webpage is smaller than the page. It is better to use % or in case of grid – fr or auto.
4)	When I was trying to make webpage less than 1440px – the left part of the page was just disappearing. To prevent it when you using grid css – just delete “justify-content: center”.

Store page
1)	I see that you’re using a lot of id(#) in you css. It’s nothing wrong but it is better to use classes for css and id for javascript.
2)	I better create form using css grid. It’s only because I think ahead about responsive design.
3)	Select the city – not working required option. Because there is already chosen an option as “Select a city”. In this way it is possible that the user won’t provide a city. It is better to use – “<option value=”” disabled selected hidden>Select your city</option>”. So then the user is required to choose the city.

You did really great job! I like your website! It looks really like a clone of that image! Most of my comments are related to responsive design. (we know that the task was only to make the clone of webpage). I  just wrote this comments to help you in future to write the website in that way – so you don’t have to write a lot of things in media queries😊

I deploy your website with some made changes. https://test-karma-boshra.netlify.app/ It’s not perfect design (because I didn’t have a time to make it look great. But I changed things about which I write above). You can check on github – changed code on my last commit. https://github.com/VitalinaKuzmenko/Test-karma-review-boshra.git

I will be very happy to explain and help you tomorrow😊

BoshraM:
@VitalinaKuzmenko 
Thank you Vitalina to look my code in this way, actually when I made Kara page I wasn't familiar with how to size my website, but now I understand what is you mean I will try to fix problem according your suggestion and I will discuss them with you in class.

PR #628, opened 2023-09-08
WM5 | Adniya Yousaf | HTML-CSS Module Project | Store Pg | Week 2

adniyaYousaf:
# Your Details

- Your Name:Adniya Yousaf
- Your City:Birmingham
- Your Slack Name:adniyaYousaf
- 
- What did you find hard?
- creating custom checkbox and radio buttons


index.html
```diff
@@ -1,19 +1,95 @@
...
+          <img
+            class="article__img"
+            src="/img/icon-refill.svg"
+            alt="icon-refill"
+          />
+          <p class="article__description">Pay as you go</p>
+        </article>
+      </section>
+      <section class="extended">
+        <div class="extended-img"></div>
```
aishaathmanlali, line 69: Can you explain the difference between div and section tags, other than their semantic values? If there is no significant difference and using semantic tags makes it more browser-friendly, should we always use section tags instead of div tags? I am seeking advice on best practices for this issue.

PR #645, opened 2023-09-29
NW6|Fidaa Bashir|HTML-CSS|Module-Project|week 1

Fidaabashir89:
# Your Details

- Your Name: Fidaa Bashir
- Your City: Manchester
- Your Slack Name: Fidaabashir89

# Homework Details

- Module: HTML-CSS-Module project-Karama project.
- Week: 1

# Notes

- What did you find easy? writing the html

- What did you find hard? to get the right size for images, and also found that to code a copy page is more difficult than to code new one ☝️ 

- What do you still not understand? I'm still need to understand grid, width and height measurements.

- Any other notes?

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,5 +1,4 @@
-
- /* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
+/* We are using the 'Roboto' font from Google. This has already been added to your HTML header */
 
 body {
     font-family: 'Roboto', sans-serif;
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -16,4 +15,107 @@ body {
...
+ nav{
+    display: flex;
+    width: 100%;
+ }
+.navbar{
+    display: flex;
+    list-style: none;
+    font-size: normal;
+    color: lightgrey;
+    padding-left: 8rem;
```
Ara225, line 43: The header is more to the left than it should be, should be right aligned as shown in the picture of the sample site
image

css/style.css
```diff
@@ -16,4 +15,107 @@ body {
...
+
+.karma-logo{
+    max-height: 40px;
+    width: auto;
+    padding-right: 10rem;
+}
+.meet{
+    font-weight: bold;
+}
+.Store{
```
Ara225, line 60: This is a small niggle, but it's worth thinking about. You're using a captial letter in this classname, and you've done it nowhere else. This can be annoying as CSS is case senstive so easy to make a mistake. It's up to you though what you want to do

css/style.css
```diff
@@ -16,4 +15,107 @@ body {
...
+    color: white;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
+}
+.learn-more{
+    background-color: rgb(235, 114, 66);
+    border-radius: 10px;
+    padding: 10px 20px;
+    margin-bottom: 5rem;
```
Ara225, line 78: This Button is a bit off I'm afraid. Has a white border and the border-radius is too big so it's more rounded than it outght to be

css/style.css
```diff
@@ -16,4 +15,107 @@ body {
...
+    font-size: 40px;
+}
+.p1{
+    margin-bottom: 1rem;
+}
+.footer-p{
+    font-weight: bold;
+}
+.copyrights{
+    color: grey;
```
Ara225, line 119: THe test on this is very small and almost unreadable for me so would make a bit bigger

css/style.css
```diff
@@ -16,4 +15,107 @@ body {
...
+}
+.line{
+    color:#DCDCDC;
+}
+.icon-footer1{
+    max-height: 20px;
+    justify-content: center;
+    border: solid 1px rgb(224, 229, 229);
+    border-radius: 15px;
+    margin-left: 30px;
```
Ara225, line 101: In the picture, the icons had a bit more space around them - they weren't touching the border

index.html
```diff
@@ -1,19 +1,67 @@
...
+    <link rel="shortcut icon" type="image/x-icon" href="favicon.ico">
 </head>
+
 <body>
 
     <!-- Add your HTML markup here -->
     <!-- Remember: Use semantic HTML tags like <header>, <main>, <nav>, <footer>, <section> etc -->
     <!-- All the images you need are in the 'img' folder -->
-
+    <header class="header"><img class="karma-logo" src="./img/karma-logo.svg" alt="karma-logo">
```
Ara225, line 19: No massive complants on the HTML here:  good use of sematic HTML,  tags have alt attrs, would say that the naming convensions could be better: nav-liwould maybe make more sense as nav-item but this is about developing your style :)

Ara225 (commented):
Hiya,
Good work, nearly there :) Copying is 100% harder work than writing from scratch but a good practice. Do you want to discuss grid and hight/width in more detail?

PR #646, opened 2023-09-29
NW6 | AREEB-SATTAR | HTML-CSS | MODULE-PROJECT | WEEK-1

areebsattar:

# Your Details

- Your Name: Areeb Sattar
- Your City: Manchester
- Your Slack Name: areebsattar

# Homework Details

- Module: HTML-CSS
- Week: 1

# Notes

- What did you find easy?
 Writing some of the HTML, 
- What did you find hard?
Found writing semantic HTML a bit tough and styling few things 
 When is it best to use relative and absolute values
- Any other notes?
N/a

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -16,4 +16,9 @@ body {
  * - When using Flexbox, remember the items you want to move around need to be inside a parent container set to 'display: flex'
  */
 
-
+.container {
+    display: grid;
+    gap: 5px;
+    grid-template-columns: repeat(12, 1fr);
+    grid-template-rows: 50px 300px 50px;
+}
--- a/css/style.css
+++ b/css/style.css
@@ -9,7 +9,7 @@ body {
  * Add your custom styles below
  * 
  * Remember: 
- * - Be organised, use comments and separate your styles into meaningful chunks
+ * - Be organized, use comments and separate your styles into meaningful chunks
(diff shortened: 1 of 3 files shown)


JayMayer:
I love what I see here Areeb. You've demonstrated a great range of HTML/CSS super powers, with advanced CSS selectors and excellent semantic HTML.

I particularly like how you use a nested flexbox for the Karma icons on the homepage. And your order form looks great, they are notoriously difficult to style - made even better by using HTML5 validations such as `required`.

Just a few pointers from me:
1. In a few places there is a class on a HTML element that is the same as the element type (e.g., `<header class="header">`). Feel free to ditch the class and just reference the element type directly from CSS (`header` instead of `.header`). This should make your code easier for other developers to work with
2. There are some slight deviations from the proposed design, for example the menu links. They're very slight, but I can guarantee the designers of the page would notice 😄

To help your uncertainty on when to use absolute/relative units: I can't give any solid answers, but as a rule of thumb I would prefer relative units wherever possible. This helps to make the site as responsive as possible. You need to be disciplined (lots of nested relative units can get confusing), but on the whole you can use relative units to make sizes that scale beautifully, whether that is based on the screen size or the user's preferred font size.

areebsattar:
Thank you for the feedback Jay, I will keep these pointers in mind for future

PR #660, opened 2024-10-04
HTML CSS Module Project 

SyedArslanHaider:
# Your Details

- Your Name:SYed Arslan Haider
- Your City: Barcelona
- Your Slack Name: Syed Arslan Haider

# Homework Details

- Module: HTML-CSS
- Week: 4

### DIFF
--- a/css/store.css
+++ b/css/store.css
@@ -1,216 +1,211 @@
 header {
-    height: 100px;
-    display: flex;
-    position: relative;
+  height: 100px;
+  display: flex;
+  position: relative;
 }
 header :nth-child(2) {
-    margin-left: auto;
-    gap: 50px;
-    margin-right: 30px;
+  margin-left: auto;
+  gap: 50px;
+  margin-right: 30px;
 }
 .burger-button {
-    display: none;
-    font-size: 30px;
-    cursor: pointer;
-    margin-left: auto;
-    text-decoration: none;
+  display: none;
+  font-size: 30px;
+  cursor: pointer;
+  margin-left: auto;
+  text-decoration: none;
 }
 header img {
-    width: 70px;
-    height: 50px;
-    padding: 15px;
-    margin-left: 3rem;
+  width: 70px;
+  height: 50px;
+  padding: 15px;
+  margin-left: 3rem;
 }
 .nav-header {
-    display: flex;
-    gap: 10px;
-    padding: 15px;
-    align-items: center;
+  display: flex;
+  gap: 10px;
+  padding: 15px;
+  align-items: center;
 }
-a{
-    text-decoration: none;
-    color: rgb(129, 121, 121);
+a {
+  text-decoration: none;
+  color: rgb(129, 121, 121);
 }
 
 a:hover {
-    color: brown;
-    font-weight: bold;
+  color: brown;
+  font-weight: bold;
 }
 
 .nav-header :nth-child(1) {
-    font-weight: bold;
-    color: rgb(85, 81, 81);
+  font-weight: bold;
+  color: rgb(85, 81, 81);
 }
 
-
-.container{
- display: flex;
+.container {
+  display: flex;
 }
-.element1{
- width: 50%;
- height: 100%;
- display: flex;
- flex-direction: column;
- align-items: center;
+.element1 {
+  width: 50%;
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
 }
-.element2{
-    width: 50%;
-    height: 100%;
+.element2 {
+  width: 50%;
+  height: 100%;
 }
-.element2 img{
-    width: 100%;
-    height: 100%;
-    object-fit: cover;
+.element2 img {
+  width: 100%;
+  height: 100%;
+  object-fit: cover;
 }
-h1{
-    text-align: center;
-    font-size: 50px;
-    padding: 30px;
-    margin-top: 3rem;
-    color: rgba(240, 81, 18, 0.904);
+h1 {
+  text-align: center;
+  font-size: 50px;
+  padding: 30px;
+  margin-top: 3rem;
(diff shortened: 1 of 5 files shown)


css/store.css
```diff
@@ -0,0 +1,289 @@
...
+  display: flex;
+  position: relative;
+}
+header :nth-child(2) {
+  margin-left: auto;
+  gap: 50px;
+  margin-right: 30px;
+}
+.burger-button {
+  display: none;
```
nsi88, line 8: Why is it hidden? Just work in progress? Or it's not needed for this design?

css/store.css
```diff
@@ -0,0 +1,289 @@
+header {
+  height: 100px;
+  display: flex;
+  position: relative;
```
nsi88, line 5: Did you already learn the differences between position static, relative, absolute, etc?
Why do you need relative in this case?

css/store.css
```diff
@@ -0,0 +1,289 @@
...
+}
+header :nth-child(2) {
```
nsi88: A similar feedback as in the past. nth-child in css is unreadable. Also it's problematic to use in more real life scenarios when content is added dynamically to the page.
Can you think of a way to avoid using it in here?

css/store.css
```diff
@@ -0,0 +1,289 @@
...
+}
+input[type="text"],
+select {
+  width: 100%;
+  padding: 12px;
+  border: 1px solid #ccc;
+  border-radius: 10px;
+  box-sizing: border-box;
+}
+.address11 {
```
nsi88: address11 and address12 are exactly the same. 
We can use just address.

nsi88: Again try to create reusable classes. What I wrote in another comment.
Class is "one of". For example "One of address fields"
Whenever there is a class address1 in your code, it's a sign that we need two classes in there:
1. Common for all address fields;
2. A modifier for a particular address field.

nsi88: Continuing with the address example.
Let's say you have 2 address fields. The first is required and has to be filled, the second is optional.
Instead of creating classes "address1" and "address2", a better way is to create classes "address" and "required".
This way if you later need a third address field, you don't need to edit your css. You will just reuse the "address" class.
Also maybe you will have other required fields having some similarities. Then the "required" class can be reused as well.

store.html
```diff
@@ -0,0 +1,105 @@
...
+    <div class="input-group">
+        <label for="postal-code">Postal Code</label>
+        <input type="text" id="postal-code" name="postal-code" placeholder="Enter postal code">
+    </div>
+</div>
+
+ <legend>Select a Color:</legend>
+<div class="radio-group">
+    <div class="radio">
+        <input type="radio" id="orange" name="orange1" value="orange">
```
nsi88: Only one color can be selected, right?
You need to use the same name for all the color inputs. "color" for example

nsi88 (commented):
The page looks really nice. Well done.
Left some comments. Feel free to ignore them if not enough time. They are just recommendations.

If you will be answering to my code comments, pls try to answer to the particular comments. I mean, not by a single big answer in the end. It's hard to find in the big answer what part is related to what)

Also please write your own comments in the PR whenever you want me to pay attention to a particular place. Maybe you have difficulties in there or you just feel proud and want to share the solution

css/store.css
nsi88: A bit more on the "required" class for address.
You could add "*" in css. It would be something like this:
```
.required::after {
  content: '*';
}
```
Haven't tested this code. Please tell if you will try and it doesn't work. Can look into it.

nsi88: "::after" above is a so called "pseudoelement" You can read more in here for example https://developer.mozilla.org/en-US/docs/Web/CSS/::after

css/store.css
SyedArslanHaider: Nice Question!
"The burger button is hidden because it's unnecessary when the page is displayed in the web view (for larger screens). I activate the burger button for mobile displays using media queries. This way, it's only visible and usable when the page is viewed on a mobile device."

css/store.css
SyedArslanHaider: "I used position: relative on the header because I want to reposition the navigation when applying media queries for mobile devices. On mobile, the navigation (header nav) is set to position: absolute, allowing me to control its position using the top, bottom, left, and right properties. By setting the parent (header) to position: relative, I ensure that the absolutely positioned navigation is placed relative to the header, making it easier to position the nav exactly where needed for mobile layouts."

css/store.css
SyedArslanHaider: "It's a bad habit of mine to use nth-child without fully considering the potential issues, but I'll make sure to avoid it in the future and look for more reliable solutions."

css/store.css
SyedArslanHaider: "Yes, it's a good approach to use just one class instead of multiple. I completely agree with you, and I'll work on resolving it. Thank you so much for your support!"

store.html
SyedArslanHaider: "Right, I will correct it "

css/store.css
SyedArslanHaider: > A bit more on the "required" class for address. You could add "*" in css. It would be something like this:
> 
> ```
> .required::after {
>   content: '*';
> }
> ```
> 
> Haven't tested this code. Please tell if you will try and it doesn't work. Can look into it.

"I used the .quote class to automatically add quotation marks before and after the content, and it works well. I’ve tested it, and everything seems fine."

css/store.css
nsi88: Makes sense. Nice one

nsi88:
It all looks good to me Syed. Well done man.
Pleas tell if there are places in the code where you want my attention.
Otherwise I review a bit selectively. Not every line.

